PR #26, opened 2024-05-21
feat(friendshipper): vendor ServerTable

cynful:
Add Servers List
Screenshot 2024-05-28 at 12 27 13 PM

### DIFF
--- a/friendshipper/src/lib/components/servers/ServerModal.svelte
+++ b/friendshipper/src/lib/components/servers/ServerModal.svelte
@@ -1,14 +1,15 @@
 <script lang="ts">
 	import { Button, Helper, Input, Label, Modal, Select, Toggle } from 'flowbite-svelte';
 	import { emit } from '@tauri-apps/api/event';
+	import { get } from 'svelte/store';
 	import { launchServer } from '$lib/gameServers';
-	import type { ArtifactEntry } from '$lib/types';
-	import { activeProjectConfig } from '$lib/stores';
+	import type { ArtifactEntry, Nullable } from '$lib/types';
+	import { activeProjectConfig, builtCommits, selectedCommit, workflowMap } from '$lib/stores';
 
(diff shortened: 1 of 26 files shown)


friendshipper/src/lib/components/servers/ServerModal.svelte
```diff
@@ -56,18 +64,57 @@
...
 <Modal
 	size="xs"
-	defaultClass="bg-secondary-700 dark:bg-space-900 overflow-y-auto"
+	defaultClass="bg-secondary-700 dark:bg-space-900 overflow-y-auto overflow-x-hidden"
 	bodyClass="!border-t-0"
 	bind:open={showModal}
+	on:open={() => {
+		selected = initialEntry !== null ? initialEntry : get(selectedCommit);
+		console.log(initialEntry);
+		console.log(selected);
```
rudoi, line 92: hmmm you shouldn't be able to commit `console.log`! :O

friendshipper/src/routes/servers/+page.svelte
```diff
@@ -0,0 +1,131 @@
...
+
+		try {
+			await updateServers();
+			selected = get(selectedCommit);
+		} catch (e) {
+			await emit('error', e);
+		}
+	};
+
+	const handleSyncClient = async (entry: Nullable<ArtifactEntry>, server: GameServerResult) => {
```
rudoi, line 46: i think this `handleSyncClient` may be unnecessary. it's used in handleAutoLaunch below, but it's not entirely clear to me why that autolaunch can't be moved into the modal.

rudoi (dismissed):
lgtm! i've seen this work. just a few comments where a few things might be worthy of a dedup, but i'm ok with shipping either way 🖖🏻 thanks for getting this in!

rudoi (approved):

PR #51, opened 2024-06-06
Updated CharacterMetadata for DCC tool needs

mkinzelman-believer:
Added metadata fields to track rigs, animation directories, and meshes.


yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
rudoi, line 1: i think you may have `yarn`'d in root without realizing it. this file can go away :D

rudoi (approved):
LGTM, i tagged this for CI so if that passes let's do it!

yarn.lock
mkinzelman-believer: Deleted

mkinzelman-believer:
Edited a couple strings in CharacterCard.svelte

rudoi (approved):

PR #53, opened 2024-06-08
feat(core): one-file-per-actor friendly names

rdunnington:
* New OFPA name cache that tries to do a web request if it finds a running UnrealEditor.exe, or falls back to the commandlet if it fails.
* OFPA names are stored in File::display_name
* Friendshipper Submit UI now shows display_name if it is set instead of the path


friendshipper/src-tauri/src/repo/operations/pull.rs
```diff
@@ -30,9 +31,11 @@ use super::{DownloadDllsOp, RepoStatusRef, StatusOp};
 #[derive(Clone)]
 pub struct PullOp {
     pub app_config: AppConfigRef,
-    pub uproject_path_relative: String,
+    pub repo_config: RepoConfigRef,
+    pub ofpa_cache: unreal::OFPANameCacheRef,
+    // pub uproject_path_relative: String,
     pub repo_status: RepoStatusRef,
-    pub trunk_branch: String,
+    // pub trunk_branch: String,
```
rudoi, line 38: can we get rid of these if we don't need em?

friendshipper/src-tauri/src/system/unreal.rs
```diff
@@ -1,11 +1,261 @@
...
+use tracing::error;
 use tracing::info;
+use tracing::warn;
+
+#[cfg(windows)]
+use anyhow::bail;
+#[cfg(windows)]
+use ethos_core::CREATE_NO_WINDOW;
+#[cfg(windows)]
+use std::path::PathBuf;
```
rudoi, line 21: just in case, for future you:

```rust
#[cfg(windows)]
use {
  anyhow::bail,
  ethos_core::CREATE_NO_WINDOW;
  std::path::PathBuf,
}
```

it doesn't actually save any lines tho lmao

core/ui/src/lib/components/repo/ModifiedFilesCard.svelte
```diff
@@ -316,7 +328,12 @@
 							>Deleted
 						</Tooltip>
 					{/if}
-					<p class="font-bold {getFileTextClass(file)}">{file.path}</p>
+					<p class="font-bold {getFileTextClass(file)}">{getFileDisplayString(file)}</p>
+					{#if file.displayName !== ''}
+						<Tooltip class="w-auto bg-secondary-600 dark:bg-space-800 font-semibold shadow-2xl">
+							{file.path}
+						</Tooltip>
```
rudoi, line 335: this tooltip, by nature of being placed inside this component, is going to be constraint to the perimeter of the component itself. so when f11r is small, it'll be totally cut off. 

i think this is shippable as-is, but folks will likely have to increase their f11r size to view the entire tooltip.

rudoi (approved):
totally sweet, nice work!

friendshipper/src-tauri/src/repo/operations/pull.rs
rdunnington: oops ty

friendshipper/src-tauri/src/system/unreal.rs
rdunnington: Oh that's right, thanks. I forgot the syntax for this use block

core/ui/src/lib/components/repo/ModifiedFilesCard.svelte
rdunnington: Yeah I think it's consistent with the rest of the app as well, unfortunately

PR #54, opened 2024-06-10
chore(friendshipper): switch friendshipper to use github releases for updating

rudoi:


friendshipper/src-tauri/src/system/update.rs
```diff
@@ -1,107 +1,161 @@
...
+                    let mut response = client
+                        .get(format!(
+                            "https://api.github.com/repos/{}/{}/releases/assets/{}",
+                            REPO_OWNER, REPO_NAME, asset.id
+                        ))
+                        .header("Authorization", format!("Bearer {}", token))
+                        .header("Accept", "application/octet-stream")
+                        .header("User-Agent", APP_NAME)
+                        .send()
+                        .await?;
```
rdunnington, line 121: I wonder if it's possible to get a callback while the stream is going to track overall progress so we could display a modal. Maybe later though.

rdunnington (dismissed):

rdunnington (approved):

PR #63, opened 2024-06-11
chore(core): plumb OFPA name changes to locks page

rdunnington:


friendshipper/src-tauri/src/repo/operations/locks/verify.rs
```diff
@@ -1,21 +1,79 @@
...
+        .load_engine_path_from_repo(&state.repo_config.read())
+        .unwrap_or_default();
+
+    let ofpa_names = OFPANameCache::get_names(
+        state.ofpa_cache.clone(),
+        &PathBuf::from(repo_path),
+        &uproject_path,
+        &engine_path,
+        &combined_paths,
+        CanUseCommandlet::FallbackOnly,
```
rudoi, line 59: This is totally shippable as-is, but this makes me wish for some kind of engine interface that's just like `Engine::sanitize_paths(Vec<String>)`.

I'll think about this as I plot out supporting different engines in f11r.

rudoi (approved):
o ya

friendshipper/src-tauri/src/repo/operations/locks/verify.rs
rdunnington: To be fair, it's almost there. We would just need to have another interface in `AppState` that switches between implementations for an engine based on a config setting for the project. Then this code could just live in the unreal version.

rudoi: For sure - this is a pattern we need to figure out for a bunch of different aspects of the tool. For example, we likely want to support non-AWS s3-like backends or Kubernetes clusters that aren't AWS EKS. Support means to me just having an interface someone else can PR implementations of.

PR #70, opened 2024-06-13
fix(friendshipper): use PAT for octocrab in update

chris-believer:
I missed the octocrab token usage in the last PR. :face_holding_back_tears:


rudoi (approved):
LGTM! wow github mobile app pretty good

PR #81, opened 2024-06-19
chore(friendshipper): introduce engine provider

rudoi:
This is the first little shot at creating an abstraction for engine interactions. We'll need a bunch more methods in the trait, but this is a start.

### DIFF
--- a/friendshipper/src-tauri/src/auth/router.rs
+++ b/friendshipper/src-tauri/src/auth/router.rs
@@ -1,25 +1,29 @@
-use std::sync::Arc;
-
 use anyhow::{anyhow, Result};
 use axum::extract::State;
 use axum::routing::{get, post};
-use axum::{debug_handler, Json, Router};
+use axum::{Json, Router};
 use ethos_core::AWSClient;
 
+use crate::engine::EngineProvider;
 use crate::APP_NAME;
 use ethos_core::clients::aws::ensure_aws_client;
 use ethos_core::types::errors::CoreError;
 
 use crate::state::AppState;
 
-pub fn router(shared_state: Arc<AppState>) -> Router {
+pub fn router<T>() -> Router<AppState<T>>
+where
+    T: EngineProvider,
+{
     Router::new()
         .route("/status", get(get_status))
         .route("/refresh", post(refresh_aws_credentials))
-        .with_state(shared_state)
 }
 
-async fn get_status(State(state): State<Arc<AppState>>) -> Json<bool> {
+async fn get_status<T>(State(state): State<AppState<T>>) -> Json<bool>
+where
+    T: EngineProvider,
+{
     let aws_client = match state.aws_client.read().await.clone() {
         Some(client) => client,
         None => {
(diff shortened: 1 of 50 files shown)


friendshipper/src-tauri/src/auth/router.rs
```diff
@@ -1,25 +1,29 @@
...
 use crate::APP_NAME;
 use ethos_core::clients::aws::ensure_aws_client;
 use ethos_core::types::errors::CoreError;
 
 use crate::state::AppState;
 
-pub fn router(shared_state: Arc<AppState>) -> Router {
+pub fn router<T>() -> Router<AppState<T>>
+where
+    T: EngineProvider,
```
rudoi, line 16: Just commenting on the first instance of this.

In reality we should only need the type args on handlers that need to make method args. Axum does allow for extracting substates, so we could have something like:

```rust
pub type AppState<T> {
  pub config: ConfigState,
  pub engine: Arc<T>
}
```

and then most of the handlers can be like:

```rust
async fn foo(State(config): State<ConfigState>) -> Result;
```

while others can be
```rust
async fn foo_with_engine<T>(State(config): State<ConfigState>, State(engine): State<T>) -> Result;
```

Axum will help with these extractions, but in order to get these substates we need to create routers that group handlers by which states they require. 

That said, I don't have a _huge_ problem with putting these type constraints everywhere, but it is tedious especially when you don't actually need any engine methods.

rudoi: I'm also trying to imagine how this pattern will expand when we have storage provider traits and kubernetes client provider traits, etc.

chris-believer (dismissed):
:icecream: Looking good to me!

friendshipper/src-tauri/src/repo/operations/download_dlls.rs
```diff
@@ -49,14 +48,16 @@ pub struct DownloadDllsOp {
     pub tx: Sender<LongtailMsg>,
     pub aws_client: AWSClient,
     pub artifact_prefix: String,
+    pub engine: Arc<T>,
```
chris-believer, line 51: Is this Arc needed? The trait requires Clone/Send/Sync/`static, so I would think it would be directly accessible?

rudoi: it is, you're totally right, good catch

friendshipper/src-tauri/src/engine/provider.rs
```diff
@@ -0,0 +1,57 @@
...
+    ) -> Vec<String>;
+
+    fn launch(&self, path: PathBuf, args: Vec<String>) -> Result<Option<Child>> {
+        if cfg!(windows) {
+            // Find the first executable in the passed path that ends with Client
+            for file in path.read_dir().context("Could not read launch directory")? {
+                let file = file.context("Invalid file")?;
+                let name = file.file_name();
+                let name = name.to_str().context("Invalid launch filename")?;
+                if name.ends_with("Client.exe") {
```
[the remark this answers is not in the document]
rudoi: easy enough to fix, so it's in!

chris-believer (approved):
:smiley: :bowing_man:

PR #84, opened 2024-06-20
fix(core): add --no-rebase when ff-only is selected for push

rudoi:


chris-believer (approved):
:+1:  being explicit is awesome!

PR #88, opened 2024-06-21
fix(friendshipper): more error handling

rdunnington:
* Handle an uncaught exception in the frontend related to null commit data
* Handle an invalid unwrap() when fetching artifacts from s3. Just return an empty list in this case to the caller and log the error.

### DIFF
--- a/friendshipper/src/lib/components/home/ContributorLayout.svelte
+++ b/friendshipper/src/lib/components/home/ContributorLayout.svelte
@@ -23,6 +23,7 @@
 		GameServerResult,
 		MergeQueue,
 		MergeQueueEntry,
+		Commit,
 		Playtest,
 		SyncClientRequest
 	} from '$lib/types';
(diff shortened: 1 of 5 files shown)


friendshipper/src/lib/components/home/ContributorLayout.svelte
```diff
@@ -221,6 +221,18 @@
 		return 'No message';
 	};
 
+	const getCommitAuthor = (commit): string => {
+		if (commit != null) {
+			if (commit.author != null) {
+				if (commit.author.name != null) {
+					return commit.author.name as string;
```
rdunnington, line 228: I had to add this `as string` to make the linter happy, but have no idea why since all the types involved are fully typed. When I manually imported `Commit` and `CommitAuthor` the linter complained they weren't used, so I dunno.

rudoi: i think you want commit: Commit in the params there maybe?

rudoi (dismissed):
just the one comment. lgtm!!

rudoi (approved):

PR #93, opened 2024-06-26
feat(core): quick submit: stacked commits

rdunnington:
* Quick Submit now has no requirement to switch back to main when making multiple submits in a row. Users may commit as many times as desired before syncing main. This is achieved using git worktree to spin up a local copy of the repo, without inflating the LFS stub files. The worktree copy is rebased on top of latest main to ensure any changes in the local branch are smoothly merged with the changes in main, which avoids limitations in GitHub's ability to detect true conflicts when pushing commits that have already flowed through the merge queue.
* Removed PushOp as most of the logic was for pre-quick submit style pushes.


core/src/clients/git.rs
```diff
@@ -463,6 +487,52 @@ impl Git {
...
+                    should_log_stdout: false,
+                    return_complete_error: true,
+                    lfs_mode: LfsMode::Stubs,
+                },
+            )
+            .await?;
+
+        let dir_regex = Regex::new(r"^worktree (.+)").unwrap();
+        let sha_regex = Regex::new(r"^HEAD (.+)").unwrap();
+        let branch_regex = Regex::new(r"^(branch|detached)\s*(.+)?").unwrap();
```
rudoi, line 513: Generally with static expressions we've tossed them in a `lazy_static!` though i should say that rust 1.70 has a solution to this that doesn't require a dependency: https://blog.rust-lang.org/2023/06/01/Rust-1.70.0.html#oncecell-and-oncelock

Non-blocking feedback either way.

core/src/clients/git.rs
```diff
@@ -507,6 +577,10 @@ impl Git {
         // disable clone protection
         cmd.env("GIT_CLONE_PROTECTION_ACTIVE", "false");
 
+        if opts.lfs_mode == LfsMode::Stubs {
+            cmd.env("GIT_LFS_SKIP_SMUDGE", "1");
```
rudoi, line 581: Awesome. I can't help but think this new opt is going to help us with a lot of stuff in the future.

friendshipper/src/routes/source/submit/+page.svelte
```diff
@@ -87,13 +86,7 @@
...
 	$: commitMessageValid = validateCommitMessage($commitMessage);
 
-	$: canSubmit =
-		$selectedFiles.length > 0 &&
-		get(commitMessage) !== '' &&
-		commitMessageValid &&
-		!loading &&
-		activeQuickSubmitPull === undefined &&
-		$repoConfig?.trunkBranch === $repoStatus?.branch;
+	$: canSubmit = $selectedFiles.length > 0 && get(commitMessage) !== '' && commitMessageValid;
```
rudoi, line 89: This feels SO good haha.

rudoi (dismissed):
All looks good to me. Super wonderful how quickly this came together :D

chris-believer (approved):
:fire:

friendshipper/src/routes/source/submit/+page.svelte
rdunnington: haha yes!

core/src/clients/git.rs
rdunnington: Yeah I can do this

rudoi (approved):

PR #94, opened 2024-06-26
feat(friendshipper): sort servers by creationTimestamp

cynful:


core/src/types/gameserver.rs
```diff
@@ -34,6 +35,7 @@ pub struct GameServerResults {
     pub port: i32,
     pub netimgui_port: i32,
     pub version: String,
+    pub creation_timestamp: Option<Time>,
```
rudoi, line 38: hello! what is the reasoning behind this field being optional? I don't think we will ever find one of these without a creation timestamp.

rudoi (dismissed):
LGTM, very excellent. I just had the one comment around creation timestamp being optional. Otherwise, do the ship. 🦖

core/src/types/gameserver.rs
cynful: there isn't a great way to default Time? like I could probably make one thats...the beginning of posix time? but it seemed fine to leave it optional to satisfy the `Default` macros

```
// Generated from definition io.k8s.apimachinery.pkg.apis.meta.v1.Time
#[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
pub struct Time(pub crate::chrono::DateTime<crate::chrono::Utc>);
```

rudoi: I think you can initialize Time from any chrono UTC datetime. We only need a default because of that stub code, so it doesn't really matter what you put there. 

I think I'd advocate for removing Option because it also removes the need for a Nullable in the frontend.

rudoi (approved):
YEAAAAA

cynful:
resolves #41

PR #113, opened 2024-07-09
fix(friendshipper): cleanup worktree branch

rdunnington:
* Leaving this branch around can be confusing to some people, so just clean it up after the merge.

### DIFF
--- a/friendshipper/src-tauri/src/repo/operations/gh/submit.rs
+++ b/friendshipper/src-tauri/src/repo/operations/gh/submit.rs
@@ -394,16 +394,16 @@ impl SubmitOp {
 
         let worktree_branch = format!("{}-wt", f11r_branch);
 
+        let mut git_client_worktree = self.git_client.clone();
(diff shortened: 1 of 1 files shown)


friendshipper/src-tauri/src/repo/operations/gh/submit.rs
```diff
@@ -473,6 +473,14 @@ impl SubmitOp {
             }
         }
 
+        // cleanup worktree branch
+        _ = git_client_worktree
```
rudoi, line 477: til you can do this without `let`

rudoi (approved):
yes

friendshipper/src-tauri/src/repo/operations/gh/submit.rs
rdunnington: I didn't even think about it, tbh. You do it this way in zig, which I'm guessing cribbed from rust :P

PR #115, opened 2024-07-10
feat(friendshipper): generate project files on macOS and Linux

holly-macdonald:


rudoi (approved):
oh, wonderful! thanks for doing this

PR #116, opened 2024-07-12
OFPA improvements & perf

rdunnington:
Note that these changes disable the OFPA commandlet. I'm planning on bringing that back as a user-facing opt-in when desired. But in general it causes too many problems to run silently in the background (blocks syncing due to an instance of UE being open, can crash on dev machines if dlls are partially compiled, etc).


friendshipper/src-tauri/src/repo/operations/status.rs
```diff
@@ -388,31 +382,33 @@ impl StatusOp {
...
+
+        // OFPANameCache::get_names(
+        //     self.ofpa_cache.clone(),
+        //     &PathBuf::from(repo_path),
+        //     &uproject_path,
+        //     &engine_path,
+        //     &filenames,
+        //     CanUseCommandlet::FallbackOnly,
+        // )
+        // .await
```
rudoi, line 411: can we blast this comment?

rudoi (approved):
Looks great, I appreciate the attention to detail on the engine provider impl. :pizza:

PR #120, opened 2024-07-19
fix(friendshipper): commit message validation bugs

rdunnington:
* After performing a quicksubmit, the message type would get reset, but the UI still showed it as selected, leading to a state mismatch with the UI, which would cause the Quick Submit button to be disabled when it looked like everything was fine.
* For the type, scope, and message fields, the message validation was bound to the input event, which is executed _before_ the bound value was actually changed, causing the message validation to be 1-off. Switching to change/keyup events ensures the input is actually completed and the bound value is updated before commit message validation runs.

### DIFF
--- a/friendshipper/src/routes/source/submit/+page.svelte
+++ b/friendshipper/src/routes/source/submit/+page.svelte
@@ -278,7 +278,7 @@
 
 			$commitMessage = '';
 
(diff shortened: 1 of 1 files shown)


rudoi (approved):
nice finds!

PR #124, opened 2024-07-21
feat(friendshipper): add otlp endpoint capabilities

rudoi:


friendshipper/src-tauri/src/repo/operations/pull.rs
```diff
@@ -310,6 +312,7 @@ where
     }
 }
 
+#[instrument(skip(state))]
```
rudoi, line 315: There is some kind of sampling or erroring that's happening somewhere that's preventing this handler from sending any traces. At first I thought it impacted all `POST`s but that wound up not being true. I'd love another set of eyes on maybe some config I missed.

core/src/utils/tracing.rs
```diff
@@ -0,0 +1,140 @@
...
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct TraceOnRequest;
+
+impl<B> OnRequest<B> for TraceOnRequest {
+    fn on_request(&mut self, _request: &Request<B>, _span: &Span) {
+        tracing::event!(
+            tracing::Level::INFO,
```
rudoi, line 64: all the events in here could easily go `DEBUG`

core/src/clients/git.rs
```diff
@@ -611,6 +613,7 @@ impl Git {
         }
     }
 
+    #[instrument(name = "git::run_and_collect_output_internal", err)]
```
rudoi, line 616: this `err` will fire an error event when any `Result<T, E>` returns an err so long as `E` implements `fmt::Display`, which I believe `anyhow` does. 

I implemented the `err` here so we don't have to instrument it at any of the call sites (which would also duplicate a bunch of sent data i think)

friendshipper/src-tauri/src/server.rs
```diff
@@ -169,24 +167,7 @@ impl Server {
 
             let app = crate::router(&shared_state.log_path)?
                 .with_state(shared_state.clone())
-                .layer(axum::middleware::from_fn(uri_passthrough))
```
rdunnington, line 172: Why did the axum layer get removed?

core/src/clients/git.rs
chris-believer: Yeah, the `std::error::Error` trait [requires it](https://doc.rust-lang.org/std/error/trait.Error.html), so anything that's an error should impl Display. :+1:

core/src/utils/logging.rs
```diff
@@ -1,42 +1,76 @@
...
+use tracing_subscriber::util::SubscriberInitExt;
+
+pub const OTEL_TRACER_PROTOCOL: opentelemetry_otlp::Protocol = opentelemetry_otlp::Protocol::Grpc;
+pub const OTEL_TRACER_TIMEOUT: Duration = Duration::from_secs(10);
+pub const ETHOS_TRACE_EVENT_TARGET: &str = "ethos::trace";
+
+pub type OtelReloadHandle = Handle<Option<OpenTelemetryLayer<Layered<EnvFilter, Registry, Registry>, Tracer>>, Layered<EnvFilter, Registry, Registry>>;
+pub fn init(prefix: &str, app: &str) -> anyhow::Result<(PathBuf, OtelReloadHandle)> {
+    // OpenTelemetry
+    let otel_layer = match env::var("OTEL_EXPORTER_OTLP_ENDPOINT") {
```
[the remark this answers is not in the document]
rudoi: I'll add a comment - this is only for local dev. The real variables come from `DynamicConfig`.

friendshipper/src-tauri/src/server.rs
rudoi: we get this behavior for free from the new layer (the uri is recorded in every span)

chris-believer (approved):

PR #127, opened 2024-07-22
Fail submitting unlocked lockable files

rdunnington:


rudoi (approved):
yeah this looks fine. i see what you're saying about wanting to make a UI pass, but let's save that for next time.

rudoi (approved):

PR #134, opened 2024-07-23
fix(core): silence git config --get error spam

rdunnington:
* If a git config variable is unset when you --get it, the process exits with error code 1, but this is a benign error that can be swallowed, and the caller can emit an error if the output isn't what they expected.


rudoi (approved):
yep!

PR #138, opened 2024-07-26
chore(friendshipper): Timestamp in System > Logs

blvr-kbrooks:
- Timestamps in System Logs now have formatting based on the System Locale.


rudoi (approved):
nailed it

PR #142, opened 2024-07-26
perf(friendshipper): revert use pathspec-from-file

rdunnington:
* The LFS post-checkout hook updates the files given, but then, if the lfs.setlockablereadonly config is true, which is on by default, it takes the hook opportunity to check all files in the repo to ensure they are properly conforming to the readonly flag. However, this process is very slow for repos with large numbers of LFS files (such as Unreal projects!). One test showed to check ~30k files took ~8s. This behavior compounds with the previous chunking behavior we were doing where reverting large amounts of files would split them up into chunks of 50 to avoid commandline limits. So reverting ~500 files (common when working with Unreal OFPA files) would be bloated by overhead of ~8s for every chunk, making the whole revert operation take minutes.
* These changes try to use the --pathspec-from-file flag, which writes all the pathspec files to a tempfile and passes it to git that way, which keeps the total number of git operations to 1 instead of n / 50.

### DIFF
--- a/friendshipper/src-tauri/src/repo/operations/revert.rs
+++ b/friendshipper/src-tauri/src/repo/operations/revert.rs
@@ -1,20 +1,20 @@
-use std::fs;
-
 use anyhow::bail;
 use axum::extract::State;
 use axum::{async_trait, Json};
+use std::fs;
+use std::io::Write;
 use tokio::sync::oneshot::error::RecvError;
+use tracing::error;
 use tracing::info;
 
 use crate::engine::EngineProvider;
+use crate::state::AppState;
 use ethos_core::clients::git;
 use ethos_core::operations::LockOp;
 use ethos_core::types::errors::CoreError;
 use ethos_core::types::locks::ForceUnlock;
 use ethos_core::types::repo::RevertFilesRequest;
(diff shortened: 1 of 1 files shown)


friendshipper/src-tauri/src/repo/operations/revert.rs
```diff
@@ -39,20 +39,101 @@ where
...
+                    error!(
+                        "Revert files with listfile failed, falling back to chunked batches: {}",
+                        e
+                    )
+                }
+            }
         }
 
-        self.git_client.run(&args, git::Opts::default()).await?;
+        for chunk in self.files.chunks(50) {
```
rudoi, line 62: do you still wanna chunk even with the listfile bit?

rdunnington: I was thinking if the listfile fails for any reason, it can fall back to the chunking code. But I suspect most of the time it won't need to chunk.

rudoi: Ah, okay, this is a fallback, cool

rudoi (approved):

PR #151, opened 2024-07-31
feat(friendshipper): move fetching and maintenance tasks into background

rudoi:
This makes all fetching on demand by polling the github API for the latest commit on the trunk.

### DIFF
--- a/core/src/clients/git_maintenance_runner.rs
+++ b/core/src/clients/git_maintenance_runner.rs
@@ -1,0 +1,88 @@
+use crate::clients::git::{Git, ShouldPrune};
+use std::path::PathBuf;
+use std::sync::mpsc::Sender as STDSender;
+use std::time::Duration;
+use tracing::{error, info};
+
+pub struct GitMaintenanceRunner {
+    git: Git,
+    config: MaintenanceConfig,
+}
+
+struct MaintenanceConfig {
+    fetch_interval: Duration,
+    maintenance_interval: Duration,
+}
+
+impl Default for MaintenanceConfig {
+    fn default() -> Self {
+        MaintenanceConfig {
+            fetch_interval: Duration::from_secs(30),
+            maintenance_interval: Duration::from_secs(3600),
+        }
+    }
+}
+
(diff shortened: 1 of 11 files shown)


core/src/clients/github_poller.rs
```diff
@@ -0,0 +1,73 @@
...
+        })
+    }
+
+    // poll and take an async fn callback
+    pub async fn poll(&mut self, interval: Duration) -> anyhow::Result<()> {
+        loop {
+            let commit = self
+                .client
+                .get_latest_commit(&self.owner, &self.repo, &self.branch)
+                .await?;
```
rudoi, line 56: I feel like I can't tell if this is actually any better than just running `git fetch`. Maybe it's better simply by nature of not messing with the repo all the time.

rudoi: Like why wouldn't we run `git fetch` every 5 seconds?

rdunnington: If `git fetch` creates the `index.lock` file, then this change is good since it's really annoying as a developer to run index-locking git operations in the repo directly only to have it error out that something else is locking the repo.

rudoi: `fetch` doesn't touch index.lock - chris and i have been discussing and we think there's a better way than this

rudoi: I'm gunna get rid of the github API stuff and instead set up background tasks for fetching and repo maintenance. i think we can get what we want that way.

core/src/clients/git_maintenance_runner.rs
```diff
@@ -0,0 +1,88 @@
...
+        self.config.fetch_interval = interval;
+        self
+    }
+
+    pub fn with_maintenance_interval(mut self, interval: Duration) -> Self {
+        self.config.maintenance_interval = interval;
+        self
+    }
+
+    pub async fn run(&mut self) -> anyhow::Result<()> {
```
rudoi, line 45: no mut needed here

PR #152, opened 2024-07-31
feat(friendshipper): implement playtest owner annotation

rudoi:


core/src/clients/git.rs
```diff
@@ -208,13 +208,21 @@ impl Git {
     pub async fn fetch(&self, prune: ShouldPrune) -> anyhow::Result<()> {
         if prune == ShouldPrune::Yes {
             self.run(
-                &["fetch", "--prune", "--no-auto-maintenance"],
+                &[
+                    "fetch",
+                    "--prune",
+                    "--no-auto-maintenance",
+                    "--show-forced-updates",
```
rudoi, line 215: added this arg to get rid of some log spam

chris-believer (approved):
lgtm!

chris-believer:
We should do something similar for GameServers at some point.

PR #153, opened 2024-08-01
feat(friendshipper): lock all button on submits view

sylviacx:


sylviacx:
Uploading safe_lock_all.mp4…

sylviacx:
https://github.com/user-attachments/assets/d6f06576-d553-4bd6-b770-ad417d090efa

core/src/tauri/command.rs
```diff
@@ -358,6 +360,77 @@ pub async fn rebase(state: tauri::State<'_, State>) -> Result<(), TauriError> {
...
+                    message: format!(
+                        "Failed to lock file: {} is locked by {}",
+                        file.path, file.locked_by
+                    ),
+                });
+            } else {
+                // log an error but continue execution
+                already_locked = true;
+                warn!(
+                    "Failed to lock file: {} is locked by {}",
```
rudoi, line 409: How about a message like "Skipping file"? That "Failed" might alarm people when they read the logs.

friendshipper/src/routes/source/submit/+page.svelte
```diff
@@ -363,6 +364,34 @@
...
+		} catch (e) {
+			await emit('error', e);
+		}
+		loading = false;
+	};
+
+	const handleLockSelected = async () => {
+		loading = true;
+		try {
+			if (await acquireLocks($selectedFiles, false)) {
```
rudoi, line 380: do we ever use "true" here? If not, i wonder if the boolean return from acquireLocks is necessary.

rudoi (changes requested):
Just a few questions! Looks real good overall!

friendshipper/src/routes/source/submit/+page.svelte
sylviacx: not for now so I can definitely leave it out and add it back in when we actually do need it

rudoi: this `handleLockSelected` will only be used by this component so i think that makes sense :+1:

core/src/tauri/command.rs
```diff
@@ -358,6 +360,77 @@ pub async fn rebase(state: tauri::State<'_, State>) -> Result<(), TauriError> {
...
+            return Err(TauriError {
+                message: format!("Failed to get github username: {}", err),
+            });
+        }
+    };
+
+    // only lock files that are unlocked
+    let mut paths: Vec<String> = Vec::new();
+    let mut already_locked = false;
+    for file in files {
```
rdunnington, line 393: We'll need to add one more failure case here, which is when the file has been modified upstream *by a different user*, we shouldn't let the user lock the file either. This will prevent cases where if the file has already been modified on latest `main`, the user won't be able to get their branch into a conflicted state.

The "by a different user" part is important too, because Quick Submit lets you submit files to `main`, but we wouldn't want to block you from submitting the file again because it technically doesn't conflict.

core/src/tauri/command.rs
```diff
@@ -358,6 +360,77 @@ pub async fn rebase(state: tauri::State<'_, State>) -> Result<(), TauriError> {
...
+        Ok(res) => res.username,
+        Err(err) => {
+            return Err(TauriError {
+                message: format!("Failed to get github username: {}", err),
+            });
+        }
+    };
+
+    // only lock files that are unlocked
+    let mut paths: Vec<String> = Vec::new();
```
rdunnington, line 391: We should move this defensive logic into the LockRequest, ideally even if a lock request comes from another source, such as Unreal, it would perform all the same checks.

core/src/tauri/command.rs
sylviacx: acquire_locks doesn't let you take a lock from someone else, it only emits warnings that it was unable to lock a file because someone else already has a lock on it. It should skip re-locking files you already have checked out and skip locking files that are owned by different users. Is this the behavior you're talking about, or do you mean if any of the requested files are already locked by someone else, we shouldn't lock anything at all?

rudoi: So if you're not up-to-date on `main`, there is a chance the file you're trying to lock is NOT locked, but IS modified in between your current commit and `main`. You should not be able to lock the file until you've pulled latest, so trying to lock it here should warn or fail depending on the failure bool.

friendshipper/src/routes/source/submit/+page.svelte
```diff
@@ -363,6 +364,34 @@
...
+		} catch (e) {
+			await emit('error', e);
+		}
+		loading = false;
+	};
+
+	const handleLockSelected = async () => {
+		loading = true;
+		try {
+			if (await acquireLocks($selectedFiles)) {
```
rudoi, line 380: ohhhhhhh im realizing i don't think i was clear on which bool i thought we didn't need - i don't think the acquire locks endpoint needs to return a `bool`. like as it sits right now, we're never going to enter the error state here because we're always skipping errors, so i think we can remove some code here.

rudoi (approved):

PR #155, opened 2024-08-01
Push status to engine

rdunnington:


friendshipper/src-tauri/src/engine/unreal/provider.rs
```diff
@@ -56,6 +47,15 @@ impl EngineProvider for UnrealEngineProvider {
         }
     }
 
+    async fn send_status_update(&self, status: &RepoStatus) {
```
rudoi, line 50: wdyt about chucking an `#[instrument(skip(self))]` on here?

rudoi (dismissed):
looks great, delightfully straightforward 🫡

friendshipper/src-tauri/src/engine/unreal/provider.rs
rdunnington: I like

friendshipper/src-tauri/src/repo/operations/status.rs
```diff
@@ -101,8 +102,15 @@ where
 {
     #[instrument(name = "StatusOp::run", skip_all)]
     pub(crate) async fn run(&self) -> anyhow::Result<RepoStatus> {
-        info!("StatusOp: running git status...");
-        let status_output = self.git_client.status(vec![]).await?;
+        info!("StatusOp: running git status and getting locks...");
```
rudoi, line 105: I think I forgot already from our discussion earlier but when do we expect the editor to call the status endpoint, if ever?

At present the StatusOp is the only thing that updates modified_upstream. The file watcher does not actually call the StatusOp, it only runs git status for each file and updates the RepoStatus directly. 

Do we need to add updating modified_upstream into the watcher too?

rdunnington: There are a couple reasons the editor still needs to call the status endpoint:
1. On startup, it needs an initial state dump.
2. The user can request a manual state refresh. I think this will be a good escape hatch if anything goes wrong and the user wants to make sure the state they're seeing is correct.

I think the only danger re: `modified_upstream` is if locks don't get updated at the same time. If the editor can see that someone has a lock on the file, or that it has been modified upstream, it won't let them check it out. Since the watcher doesn't update `modified_upstream` or locks, I think it's OK since if a file is going to be submitted, it needs a lock, and the StatusOp contains updates for both of them now.

rudoi (approved):

PR #162, opened 2024-08-02
fix(friendshipper): canonicalize path when setting readonly

rudoi:


core/src/operations.rs
```diff
@@ -283,25 +283,29 @@ impl LockOp {
...
-                                    perms.set_readonly(set_readonly);
-
-                                    if let Err(e) = std::fs::set_permissions(&absolute_path, perms)
-                                    {
-                                        error!(
-                                            "Failed to {} readonly flag for file {:?}: {}",
-                                            operation_str, &absolute_path, e
-                                        );
+                        // canonicalize path (this cleans up the path and ensures existence)
+                        if let Ok(canonical_path) = absolute_path.canonicalize() {
```
rdunnington, line 287: Do we want to handle a potential `canonicalize()` error?

rudoi:
```
This function will return an error in the following situations, but is not limited to just these cases:
path does not exist.
A non-final component in path is not a directory.
```

Eh? We're basically ignoring the Not Found case here, and I cannot imagine hitting the other case, but maybe?

rudoi: I feel like "skip any file that fails this check" is reasonable, but am open to arguments

rdunnington (approved):

core/src/operations.rs
rudoi: logged the error 🖖🏻

PR #165, opened 2024-08-03
fix(core): check if file exists before canonicalizing for read-only

sylviacx:


sylviacx:
Doesn't look like the issue comes from an incorrect path input, as at the line where we emit the error log, absolute_path is still well formed. 

Instead, if users try locking files that have been deleted off disk, canonicalize won't work. But if they're off disk, there's no need to flip the read-only flag, so we should just skip that chunk of logic all together.

rudoi (approved):

PR #166, opened 2024-08-05
fix(friendshipper): fix git hooks not working on macOS

holly-macdonald:


friendshipper/src-tauri/src/repo/operations/install_git_hooks.rs
```diff
@@ -64,6 +67,12 @@ impl Task for InstallGitHooksOp {
 
                     debug!("copying {:?} to {:?}", entry.path(), destination);
                     fs::copy(entry.path(), &destination)?;
+
+                    #[cfg(target_os = "macos")]
```
rudoi, line 71: wdyt about going for `not(target_os = "windows")` here? so we "accidentally" get linux support

rudoi (approved):
looks good! just had one comment

friendshipper/src-tauri/src/builds/router.rs
```diff
@@ -355,6 +355,7 @@ where
         let argolabels = workflow.metadata.labels.as_ref().unwrap();
         let argoannotations = workflow.metadata.annotations.as_ref().unwrap();
 
+        #[allow(clippy::manual_inspect)]
```
rdunnington, line 358: Is this from a linter suggestion?

rdunnington (approved):

friendshipper/src-tauri/src/builds/router.rs
holly-macdonald: Yeah. At first I tried fixing this by replacing the `map` whose return value we check with `inspect` as suggested; but this just caused the linter to trigger the same error on the other instances of `map` whose return value we're not using, even after I tried switching them to `inspect`

PR #167, opened 2024-08-06
feat(friendshipper): add sorting to modified files table

sylviacx:


rudoi (approved):
yeah this looks great

PR #170, opened 2024-08-06
feat(friendshipper): add button to directly open powershell from pref…

sylviacx:
…erences

### DIFF
--- a/friendshipper/src-tauri/src/system/terminal.rs
+++ b/friendshipper/src-tauri/src/system/terminal.rs
@@ -6,7 +6,7 @@ use tracing::error;
 use crate::engine::EngineProvider;
 use crate::state::AppState;
 
-#[cfg(windows)]
+#[cfg(any(target_os = "windows", target_os = "linux"))]
 pub async fn open_terminal_to_path<T>(State(_state): State<AppState<T>>, path: String)
 where
     T: EngineProvider,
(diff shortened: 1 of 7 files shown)


friendshipper/src-tauri/src/system/terminal.rs
```diff
@@ -0,0 +1,24 @@
...
+use tracing::error;
+
+use crate::engine::EngineProvider;
+use crate::state::AppState;
+
+pub async fn open_terminal_to_path<T>(State(_state): State<AppState<T>>, path: String)
+where
+    T: EngineProvider,
+{
+    if let Err(e) = Command::new("cmd")
```
rudoi, line 13: i think we'll need to guard this with `#[cfg(windows)]`, that way we can add a separate path for mac/linux etc.

friendshipper/src/lib/components/preferences/PreferencesModal.svelte
```diff
@@ -278,14 +284,22 @@
...
+						/>
+						<Tooltip class="text-sm" placement="bottom">
+							Specified folder must be a game repository.
+						</Tooltip>
+						<Button
+							class="h-8 bg-primary-700 hover:bg-primary-800 dark:bg-primary-600 hover:dark:bg-primary-700"
+							on:click={openTerminalToRepo}
+						>
+							<TerminalSolid class="w-4 h-4" color="white" />
+						</Button>
```
rudoi, line 300: How about another tooltip here like `Open terminal at repo path`? just so folks know what they're clicking :D

rudoi (changes requested):
looks great! just a few thingies

PR #177, opened 2024-08-09
fix(core): remove spaces on some setup inputs

sylviacx:
not sure if core is the right scope? it's not technically in the core package but it is a fix to both friendshipper and birdie

### DIFF
--- a/friendshipper/src/lib/components/oobe/WelcomeModal.svelte
+++ b/friendshipper/src/lib/components/oobe/WelcomeModal.svelte
@@ -74,22 +74,22 @@
 					awsSsoStartUrl !== '' &&
 					awsRoleName !== '' &&
 					artifactBucket !== '';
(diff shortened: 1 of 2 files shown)


sylviacx:
https://github.com/user-attachments/assets/4d775ef4-be4e-42b6-93fb-caf7eab82956

friendshipper/src/lib/components/oobe/WelcomeModal.svelte
```diff
@@ -74,15 +74,22 @@
 					awsSsoStartUrl !== '' &&
 					awsRoleName !== '' &&
 					artifactBucket !== '';
+				awsAccountId = awsAccountId.replaceAll(' ', '');
```
rudoi, line 77: I think string in javascript has trim() which gets rid of all leading and trailing whitespace. wdyt about using that?

PR #185, opened 2024-08-13
feat(friendshipper): add support for multiple playtest regions

rudoi:
This enables switching between regions for deploying playtests + servers. Enabled regions are enumerated as part of dynamic config. 

At present the default is still `us-west-2`.

### DIFF
--- a/friendshipper/src/lib/components/preferences/PreferencesModal.svelte
+++ b/friendshipper/src/lib/components/preferences/PreferencesModal.svelte
@@ -11,6 +11,7 @@
 		Label,
 		Modal,
 		Radio,
+		Select,
 		Spinner,
 		Toggle,
 		Tooltip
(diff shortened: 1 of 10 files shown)


friendshipper/src/lib/components/preferences/PreferencesModal.svelte
```diff
@@ -226,6 +235,20 @@
 				For engineers. Enable if you want to debug the game client locally. Increases download size.
 			</Tooltip>
 		</div>
+		{#if $dynamicConfig.playtestRegions.length > 1}
```
sylviacx, line 238: If we only had one playtest region, should we still show it in preferences so users know which region they're in?

friendshipper/src/routes/servers/+page.svelte
```diff
@@ -42,6 +42,10 @@
 		}
 	};
 
+	void listen('preferences-closed', () => {
```
sylviacx, line 45: What does this fn do, how is it used if it's not called directly on this page?

rudoi: The preferences modal itself emits a `preferences-closed` event, which this component listens to.

friendshipper/src/lib/components/preferences/PreferencesModal.svelte
rudoi: I think I could go either way. In my head if there's only one region, I don't think the info is useful for typical users.

sylviacx: gotcha, we can keep as is then

sylviacx (approved):

PR #187, opened 2024-08-13
refactor(birdie): pull birdie config.yaml out of core

cynful:
will move the friendshipper config.yaml out of core in a seperate PR

### DIFF
--- a/birdie/src-tauri/src/config/router.rs
+++ b/birdie/src-tauri/src/config/router.rs
@@ -5,7 +5,9 @@ use axum::{debug_handler, extract::State, routing::get, Json, Router};
 use tracing::info;
 
 use ethos_core::clients::github::GraphQLClient;
-use ethos_core::types::config::AppConfig;
+
+// todo: mv ethos_core::types::config::BirdieConfig to birdie::types::config::AppConfig
+use ethos_core::types::config::BirdieConfig as AppConfig;
 use ethos_core::types::config::ConfigValidationError;
 use ethos_core::types::errors::CoreError;
 
--- a/birdie/src-tauri/src/config/router.rs
+++ b/birdie/src-tauri/src/config/router.rs
@@ -5,9 +5,6 @@ use axum::{debug_handler, extract::State, routing::get, Json, Router};
 use tracing::info;
 
 use ethos_core::clients::github::GraphQLClient;
-
-// todo: mv ethos_core::types::config::BirdieConfig to birdie::types::config::AppConfig
-use ethos_core::types::config::BirdieConfig as AppConfig;
 use ethos_core::types::config::ConfigValidationError;
 use ethos_core::types::errors::CoreError;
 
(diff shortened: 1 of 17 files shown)


core/src/types/config.rs
```diff
@@ -44,14 +44,82 @@ pub struct AWSConfig {
     pub artifact_bucket_name: String,
 }
 
-#[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct AppConfig {
+pub type SharedConfigRef = Arc<RwLock<SharedConfig>>;
+
+#[derive(Clone, Debug, Serialize, Deserialize, Default)]
+pub struct SharedConfig {
```
rudoi, line 50: Can you help me understand the purpose of this struct a bit more?

I see how it's used for the RepoWorker, but am curious if maybe there's a different way we can initialize the RepoWorker. 

It feels to me like having this shared struct actually adds more code than it saves because of all the places you need to check for it. I'm also a tad bit concerned about lock contention and timing issues as we've now got a lockable inside a lockable, which isn't categorically bad but I could see us deadlocking in the future if we aren't careful.

core/src/types/config.rs
```diff
@@ -44,14 +44,82 @@ pub struct AWSConfig {
...
+            )
+            .into()),
+        }
+    }
+}
+
+pub type BirdieConfigRef = Arc<RwLock<BirdieConfig>>;
+
+#[derive(Clone, Debug, Serialize, Deserialize, Default)]
+pub struct BirdieConfig {
```
rudoi, line 104: wdyt about moving `BirdieConfig` out of `core`, renaming `AppConfig` to `FriendshipperConfig`, and moving that out of `core` too?

cynful: that is the plan, I wanna do this is a second PR
becuase there's just so much going on

core/src/types/config.rs
cynful: its just for shared fields
I do agree, the double lock mechanism does worry me

Im fine with duplicating fields, and then refactoring the functions to take args instead

core/src/types/config.rs
rudoi: totally fine to punt on that for this PR then IMO

birdie/src-tauri/src/config/router.rs
```diff
@@ -16,6 +15,7 @@ use ethos_core::types::repo::CloneRequest;
 use {crate::DEFAULT_DRIVE_MOUNT, ethos_core::utils, std::path::Path};
 
 use crate::state::AppState;
+use crate::types::config::AppConfig;
```
cynful, line 18: should this be `birdie::types::config::AppConfig` or `birdie::types::config::BirdieConfig`

for clarity I prefer `BirdieConfig`

thoughts @rudoi ?

rudoi: i agree

core/src/types/config.rs
```diff
@@ -198,15 +198,15 @@ impl AppConfig {
...
 
-    pub fn ensure_github_pat(&self) -> Result<String, CoreError> {
-        match self.github_pat.clone() {
-            Some(pat) => Ok(pat),
-            None => Err(anyhow!(
-                "GitHub PAT is not configured. Please configure it in the settings.".to_string()
-            )
-            .into()),
-        }
+pub fn ensure_github_pat(github_pat: &Option<String>) -> Result<String, CoreError> {
```
rudoi, line 203: this function is a little bit strange now, i feel like we could just remove it and at the callsites just do like 

```rs
let token = config.github_pat.ok_or(CoreError::from.......
```

rudoi (approved):
LGTM ye

PR #193, opened 2024-08-13
fix(ethos): Prevent runaway process creation

blvr-kbrooks:
- Adding check to prevent hundreds of git processes from spawning because the user's credentials expired. Only checking if several git-credential-manager instances are running simultaneously. Any more than 4 will prevent a git command from executing — hopefully this is high enough to prevent false positives, but low enough to prevent the issue.

### DIFF
--- a/core/src/clients/git.rs
+++ b/core/src/clients/git.rs
@@ -7,7 +7,7 @@ use chrono::{DateTime, Utc};
 use lazy_static::lazy_static;
 use parking_lot::RwLock;
 use regex::Regex;
-use sysinfo::{System};
+use sysinfo::{ProcessRefreshKind, System, UpdateKind};
 use tokio::io::AsyncBufReadExt;
 use tokio::io::BufReader;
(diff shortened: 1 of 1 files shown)


core/src/clients/git.rs
```diff
@@ -628,6 +629,13 @@ impl Git {
             bail!("No arguments provided to git command");
         }
 
+        let mut sys = System::new();
+        sys.refresh_processes();
```
rudoi, line 633: In the past we've found that this call can be exorbitantly expensive - up to 20 seconds even for a PC doing normal amounts of work. If you peep `process.rs`, we've found a way to filter in a way that makes this a bit cheaper:

```rs
 let mut system = sysinfo::System::new();
 let refresh_kind =
        sysinfo::ProcessRefreshKind::new().with_exe(sysinfo::UpdateKind::OnlyIfNotSet);
 system.refresh_processes_specifics(refresh_kind);

rudoi (approved):

PR #194, opened 2024-08-13
chore(friendshipper): unify snapshot behavior on submit

rudoi:
This deprecates the behavior where we lay down a temporary `.stash` directory. 

What's interesting is that we're actually able to ignore the `unable to unlink` error that's thrown when the editor is open. I noticed in my testing that the snapshot totally works. Git must be attempting to touch some metadata on the file that is irrelevant to us.

### DIFF
--- a/core/src/clients/git.rs
+++ b/core/src/clients/git.rs
@@ -407,7 +407,11 @@ impl Git {
             stash_args.push(path);
         }
 
-        self.run(&stash_args, Opts::default()).await?;
+        // In testing, we found that regularly this would throw unlink errors when the editor was
+        // open. However, the snapshot still gets created successfully, so in this particular case
+        // we can ignore the error.
+        self.run(&stash_args, Opts::new_with_ignored(&["unable to unlink"]))
+            .await?;
 
         let snapshots = self.list_snapshots().await?;
 
--- a/friendshipper/src-tauri/src/repo/operations/gh/submit.rs
+++ b/friendshipper/src-tauri/src/repo/operations/gh/submit.rs
@@ -1,7 +1,6 @@
 use anyhow::anyhow;
 use axum::extract::State;
 use axum::{async_trait, Json};
-use directories_next::ProjectDirs;
 use octocrab::models::pulls::MergeableState;
 use octocrab::Octocrab;
 use std::path::PathBuf;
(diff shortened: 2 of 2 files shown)


core/src/clients/git.rs
```diff
@@ -407,7 +407,8 @@ impl Git {
             stash_args.push(path);
         }
 
-        self.run(&stash_args, Opts::default()).await?;
+        self.run(&stash_args, Opts::new_with_ignored(&["unable to unlink"]))
```
rudoi, line 410: I actually do not know why we can get away with this.

rdunnington, line 410: This is probably worth a comment

friendshipper/src-tauri/src/repo/operations/gh/submit.rs
```diff
@@ -220,53 +219,38 @@ where
...
 
         match self.execute_internal().await {
             Ok(_) => Ok(()),
             Err(e) => {
                 // attempt to reset to original branch and restore snapshot
                 // if this fails for any reason, we should simply log, then return the original error
-                let _ = self
-                    .recover(stash_path, &current_branch, stashed_files, deleted_files)
+                let modified_files = self.repo_status.read().modified_files.clone();
+                let untracked_files = self.repo_status.read().untracked_files.clone();
```
rdunnington, line 233: I really need to delete `untracked_files` now that `File` stores the git state

friendshipper/src-tauri/src/repo/operations/gh/submit.rs
```diff
@@ -220,53 +219,38 @@ where
...
-                        warn!(
-                            "Failed to recover from error during submit operation: {}",
-                            err
-                        );
-                    });
+                {
+                    Ok(_) => {
+                        // if we successfully restored the snapshot, we should be able to reset to the original branch
+                        let branch = self.repo_status.read().branch.clone();
+                        self.git_client.hard_reset(&branch).await?;
```
rdunnington, line 248: won't this wipe out all the changes that just got restored? do we want to do this _before_ calling `restore_snapshot`?

rudoi: hahahaha yeah.....

core/src/clients/git.rs
rdunnington: this is good enough for me, ty

rdunnington (approved):

rdunnington (approved):

PR #195, opened 2024-08-13
chore(core): update repo state with updated locks after lock op

rdunnington:
* Now that we're pushing status updates to the engine, updating the repo status with the latest lock info ensures that it has more up-to-date locks info when sending status updates after a filewatch trigger.
* Moved the LockOp to be run in a task sequence to avoid racing with the status op


rudoi (approved):
yep!

PR #201, opened 2024-08-14
refactor(birdie): rm unused RepoConfig

cynful:


rudoi (approved):

birdie/src-tauri/src/repo/push.rs
```diff
@@ -158,7 +157,7 @@ pub async fn push_handler(
         files: request.files.clone(),
         git_client: state.git(),
         repo_status: state.repo_status.clone(),
-        trunk_branch: state.repo_config.read().trunk_branch.clone(),
+        trunk_branch: "main".to_string(),
```
cynful, line 160: this was the thing that I was not sure was bad or not
@rudoi

cynful: should i write a new function for git to get the remote HEAD? and call that instead of just assigning main?

rudoi: IIRC we were defaulting this to main anyway so this isn't strictly worse.

cynful: correct it is defaulted to main

I'm going to proceed as is

PR #203, opened 2024-08-14
fix(friendshipper): allow quicksubmit when aws token is expired

rdunnington:
* QuickSubmit doesn't perform any AWS operations since it skips the dll lookup when running StatusOp, so just make it optional to avoid failing the submit. This can happen when the AWS auth token expires and the user (or Friendshipper) hasn't noticed yet.


rudoi (approved):
ye im good with this

PR #213, opened 2024-08-15
fix(core): only drop 26th stash when there are greater than 25

rudoi:
If you drop a stash in the middle of the stack, all the indices change, duh! So in cases where we were trying to drop two stashes, we'd drop 1, then the 2nd one might go out of bounds.


friendshipper/src-tauri/src/builds/router.rs
```diff
@@ -355,7 +355,6 @@ where
         let argolabels = workflow.metadata.labels.as_ref().unwrap();
         let argoannotations = workflow.metadata.annotations.as_ref().unwrap();
 
-        #[allow(clippy::manual_inspect)]
```
rudoi, line 358: according to my clippy this is unnecessary... i dunno...

cynful (approved):

PR #216, opened 2024-08-15
feat(birdie): lock all button

sylviacx:
#7212272047


sylviacx:
still need terminal button, coming in next commit

birdie/src/routes/source/submit/+page.svelte
```diff
@@ -96,15 +100,46 @@
...
+		progressModalTitle = '';
 		loading = false;
 	};
 
 	onMount(() => {
 		void refreshFiles(true);
 
 		const interval = setInterval(async () => {
-			if (!submitting) {
+			if (progressModalTitle === 'Submitting') {
```
rudoi, line 142: i think maybe just `if (!loading)` here maybe? checking a string value here gives me the ick

rudoi (approved):
looks great ty!!!

PR #217, opened 2024-08-15
feat(birdie): open terminal buttons

sylviacx:
#7212272047

### DIFF
--- a/birdie/src-tauri/src/system/terminal.rs
+++ b/birdie/src-tauri/src/system/terminal.rs
@@ -24,5 +24,5 @@ pub async fn open_terminal_to_path(State(_state): State<Arc<AppState>>, path: St
 
 #[cfg(not(target_os = "windows"))]
 pub async fn open_terminal_to_path(State(_state): State<Arc<AppState>>, _path: String) {
-    error!("Open terminal not supported on linux");
+    error!("Open terminal not supported on linux or macOS");
 }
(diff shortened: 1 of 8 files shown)


birdie/src-tauri/src/system/terminal.rs
```diff
@@ -0,0 +1,22 @@
+use std::process::Command;
+use std::sync::Arc;
+
+use axum::extract::State;
+use tracing::error;
+
+use crate::state::AppState;
+
+#[cfg(any(target_os = "windows", target_os = "linux"))]
```
rudoi, line 9: powershell isn't a thing in linux so i'd recommend doing only windows here, then implementing a second function called `open_terminal_to_path` with the same signature that simply logs an error for not windows

birdie/src-tauri/src/system/terminal.rs
```diff
@@ -0,0 +1,28 @@
...
+        .arg("-NoExit")
+        .arg("-Command")
+        .arg(format!("cd {}", path))
+        .spawn()
+    {
+        error!("Error opening terminal to path: {:?}", e);
+    }
+}
+
+#[cfg(target_os = "linux")]
```
rudoi, line 25: hehe can you swap this for `#[cfg(not(target_os = "windows"))]` - for mac users

rudoi (dismissed):

rudoi (approved):

PR #223, opened 2024-08-22
feat(birdie): favorite files feature

sylviacx:
#6770525631


sylviacx:
TODO: reconcile config and files already downloaded, ability to unfavorite files

sylviacx:
TODO: reconcile paths -> directories when all files within a directory are added to lfs.fetchinclude

Cargo.toml
BelieverCoho, line 1: intended removal of these 2? :

sylviacx: not intended you're right, just fixed!

rudoi (approved):
let's go!!!!

PR #224, opened 2024-08-26
doc(friendshipper): add build and artifact docs

chris-believer:


BelieverCoho (dismissed):
Looks good to me! I think some more examples could be good, also the "Game Builds" section mention of the docker file and the entrypoint.sh could benefit from that as well as I suspect that would be a question for a person unfamiliar with them.

rudoi (approved):
looks great, ty!

PR #225, opened 2024-09-03
fix(friendshipper): allow 25 playtest groups

chris-believer:
Screenshot from 2024-09-03 12-48-34

### DIFF
--- a/friendshipper/docs/sourcecontrol.md
+++ b/friendshipper/docs/sourcecontrol.md
@@ -1,51 +1,56 @@
 # Overview
 
-Friendshipper is designed to be a one-stop-shop for all nontechnical users' source control needs. For Unreal projects, 
+Friendshipper is designed to be a one-stop-shop for all nontechnical users' source control needs. For Unreal projects,
 it essentially replaces workflows that UGS and P4V support.
 
 ### Features
 
-* Commit history
-* Syncing repo + engine/editor binaries
-* Submitting files + optional commit message formatting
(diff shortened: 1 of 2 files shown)


friendshipper/src/lib/components/playtests/PlaytestModal.svelte
```diff
@@ -256,7 +262,7 @@
 					size="sm"
 					name="minGroups"
 					min="1"
-					max="10"
+					max={$dynamicConfig.playtestDiscordChannels.length + 1}
```
rudoi, line 265: i feel like maybe we either set max to a higher const or remove it altogether - something irks me about requiring discord to playtest at all, yknow?

chris-believer: I agree, I kinda hated doing it. 

In an ideal world I'd love some reasonable triggerpoint with a warning, requiring a "Confirm" to continue anyway. (Just to underscore the resource allocation it's making.) 

I think just bumping to, say, 25 would be fine, too.

rudoi (approved):

friendshipper/docs/sourcecontrol.md
```diff
@@ -1,51 +1,56 @@
 # Overview
 
-Friendshipper is designed to be a one-stop-shop for all nontechnical users' source control needs. For Unreal projects,
```
rudoi, line 3: now that's some formatting

PR #226, opened 2024-09-03
fix(core): put autostash flag at the end

rudoi:
This is a little bit of a shot in the dark - git seems to sometimes be parsing `--autostash` as an additional ref argument, resulting in `cannot rebase onto multiple branches`.


chris-believer (approved):
Wow, crazy and great find. GitOxide will save us one day... :smile:

PR #227, opened 2024-09-03
chore(core): split out internal and input errors

rudoi:


core/src/types/errors.rs
```diff
@@ -4,21 +4,32 @@ use axum::response::{IntoResponse, Response};
...
+            }
+        }
     }
 }
 
 impl std::fmt::Display for CoreError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.0)
+        match self {
+            CoreError::Input(e) => write!(f, "{}", e),
```
chris-believer, line 30: Worth any differentiation in the string for logs, or does the error type already show up?

chris-believer (approved):
One QQ, but lgtm!

core/src/types/errors.rs
rudoi: I think it's okay this way, the big differentiation is in the status code (400 vs 500). happy to go either way though

chris-believer (approved):

PR #229, opened 2024-09-03
fix(birdie): use CoreError enums to fix build error

sylviacx:


birdie/src-tauri/src/repo/lfs.rs
```diff
@@ -113,14 +113,14 @@ pub async fn del_fetch_include(
                     info!("Successfully set lfs.fetchinclude to {}", all_paths_str);
                 }
                 Err(e) => {
-                    return Err(CoreError(anyhow!(
+                    return Err(CoreError::Input(anyhow!(
```
rudoi, line 116: i think these may all actually be `CoreError::Internal`

rudoi: if the error is due to a user input/choice - `Input`, if it's some kind of system error that may not be the user's fault - `Internal`

sylviacx: I was thinking writes/opens would fail if either the file path was invalid because repo_path is incorrect or if the file was set to read only or something. In those cases I thought it would be more of an input error, since either the user has the wrong path in birdie config or somehow the file on the user's machine is set to read only. What are your thoughts?

rudoi: oh sure, that makes sense to me! ship as is

rudoi (approved):

PR #237, opened 2024-09-05
fix(birdie): use birdie specific get_config + update_config

cynful:
During cutover, I did not update the core get_app_config and update_app_config to handle two config types (friendshipper config and birdie config)

Fixed so that birdie owns it's own get/update_config commands


rudoi (approved):
mamma mia

PR #239, opened 2024-09-05
feat(friendshipper): add copy mobile launch URL button to servers list

holly-macdonald:


rudoi (dismissed):
lgtm!

core/src/types/config.rs
```diff
@@ -106,6 +106,9 @@ pub struct AppConfig {
 
     #[serde(default)]
     pub initialized: bool,
+
+    #[serde(default, rename = "mobileURLScheme")]
+    pub mobile_url_scheme: String,
```
rudoi, line 111: if this all works the way you're intending it to right now, i'm cool with shipping it, but I do feel like this might be better off in `RepoConfig` rather than `AppConfig`. 

`RepoConfig` is a structure held in the root of the game project's repository (`friendshipper.yaml`). This way, it can be set at the project level and we won't need to ask individual users to manually set the url scheme themselves.

holly-macdonald: Makes sense to me, this can definitely be something that one person sets up for the whole team

rudoi (approved):

PR #240, opened 2024-09-05
chore(friendshipper): force a restart to re-prompt auth

rudoi:
restarting seems like the safest way to make sure we don't make a bunch of requests after login expires.


BelieverCoho (dismissed):

BelieverCoho (approved):

rudoi:
Still putting this through its paces. I will probably merge next week.

PR #244, opened 2024-09-07
fix(friendshipper): move mobile scheme into dynamic config

rudoi:
Dynamic config is best here because it allows the config to be accessible to all users, not just users who have the game repo cloned.


holly-macdonald (approved):
looks good to me!

PR #252, opened 2024-09-13
chore(core): make otel accessible during startup

rudoi:
This change allows us to start the app with otel running. For Birdie, because we don't yet have an otel mechanism, this does nothing yet. For Friendshipper, we now allow caching otel config returned from dynamic config in the local app config file. So, your very first start up will not be instrumented, but provided that there is otel config in dynamic config, subsequent startups will be instrumented.

### DIFF
--- a/friendshipper/src-tauri/src/main.rs
+++ b/friendshipper/src-tauri/src/main.rs
@@ -6,6 +6,8 @@
 use std::thread;
 
 use ethos_core::longtail::Longtail;
+use ethos_core::types::errors::CoreError;
+use friendshipper::server::Server;
 use lazy_static::lazy_static;
 use serde::Serialize;
 use tauri::api::notification::Notification;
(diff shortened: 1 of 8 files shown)


friendshipper/src-tauri/src/main.rs
```diff
@@ -76,346 +78,369 @@ fn force_window_to_front(window: Window) {
...
-
-    match utils::process::check_for_process(APP_NAME) {
-        Ok(_) => {}
-        Err(e) => {
-            error!("Failed to check for existing process: {:?}", e);
-            std::process::exit(1);
-        }
-    }
+    if let (Some(config_file), Some(config)) = Server::initialize_app_config()? {
+        let (log_path, otel_reload_handle) = match tauri::async_runtime::block_on(async {
```
rudoi, line 90: need to use tauri's async_runtime singleton here so that the tracing runtime starts within the same tokio runtime as the main application

rdunnington (approved):

PR #263, opened 2024-09-19
fix(friendshipper): no lockable read only

rudoi:


core/src/operations.rs
```diff
@@ -257,28 +257,11 @@ impl LockOp {
...
-                Ok(env_str) => {
-                    if let Ok(env_bool) = git::parse_bool_string(&env_str) {
-                        should_set_read_flag = env_bool;
-                    }
-                }
-                Err(_) => {
-                    let output = self
-                        .git_client
-                        .run_and_collect_output(
-                            &["config", "--get", "lfs.setlockablereadonly"],
```
rdunnington, line 272: Why did this code get deleted? It's checking the setting for the lockablereadonly flag.

rudoi: we ensure this setting is off on startup, so not sure it's worth checking? i could leave it in for safety but if our code doesn't intend to honor it....

rdunnington: ok yeah that makes sense

rdunnington (approved):

rdunnington (approved):

PR #276, opened 2024-09-25
feat(friendshipper): git reset --keep <sha>

cynful:
this allows users to reset repo to a previous commit and keep their local changes

### DIFF
--- a/friendshipper/src-tauri/src/repo/operations/reset.rs
+++ b/friendshipper/src-tauri/src/repo/operations/reset.rs
@@ -1,7 +1,22 @@
-use crate::engine::EngineProvider;
-use crate::state::AppState;
-use axum::extract::State;
+use anyhow::{anyhow, Context};
+use async_trait::async_trait;
+use axum::extract::{Path, State};
+
+use ethos_core::clients::aws::ensure_aws_client;
+use ethos_core::clients::git;
+use ethos_core::types::config::{RepoConfig, UProject};
 use ethos_core::types::errors::CoreError;
+use ethos_core::worker::{Task, TaskSequence};
+
+use crate::engine::EngineProvider;
+use crate::repo::operations::{DownloadDllsOp, StatusOp, UpdateEngineOp};
+use crate::state::AppState;
+use ethos_core::storage::config::Project;
+use ethos_core::storage::ArtifactStorage;
+use ethos_core::AWSClient;
+use std::path::PathBuf;
+use tokio::sync::oneshot::error::RecvError;
+use tracing::{error, info};
 
 pub async fn reset_repo<T>(State(state): State<AppState<T>>) -> Result<(), CoreError>
 where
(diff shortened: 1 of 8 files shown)


friendshipper/src-tauri/src/repo/operations/reset.rs
```diff
@@ -1,12 +1,59 @@
...
+}
+
+#[async_trait]
+impl Task for ResetToCommitOp {
+    async fn execute(&self) -> Result<(), CoreError> {
+        self.git_client
+            .run(&["reset", "--hard", &self.commit], Default::default())
+            .await
+            .map_err(CoreError::Internal)
+    }
```
rdunnington, line 31: This will also need to run a DownloadDllsOp and UpdateEngineOp as well, since the game and engine code may have changed between commits.

rdunnington (changes requested):

friendshipper/src-tauri/src/repo/operations/reset.rs
rudoi: I think you'll also need to run a StatusOp before running those downloads as I believe it's the StatusOp that sets the right commit for the DownloadDllsOp to use. I may be wrong.

cynful: @rudoi in the refresh() call we invoke get_repo_status -> status_handler (StatusOp)

friendshipper/src/routes/source/history/+page.svelte
```diff
@@ -167,6 +170,34 @@
...
 		await refresh();
 	};
+
+	const handleResetRepoToCommit = async (commit: string) => {
+		try {
+			inAsyncOperation = true;
+			asyncModalText = 'Resetting repo to commit...';
+			await resetRepoToCommit(commit);
+			selectedCommit = '';
+			await refresh();
```
cynful, line 180: @rudoi here's where the StatusOp gets called after commit change

friendshipper/src-tauri/src/repo/operations/reset.rs
rudoi: i think the reason we're specifically calling out the Ops is so that they're included in the TaskSequence that executes the reset. in its current state, the reset endpoint leaves friendshipper in a little bit of an unknown state, whereas you can use the TaskSequence to ensure that everything you need gets executed. if you do this sequencing in the frontend, there's no guarantee some other queued request won't make some kind of destructive change to the repo. does that make sense?

rdunnington (approved):
I was thinking the `ResetToCommitOp` would embed the ops necessary to make it all "correct", but one nice benefit of putting the ops in the frontend is we can give better status updates to users.

rudoi:
> I was thinking the `ResetToCommitOp` would embed the ops necessary to make it all "correct", but one nice benefit of putting the ops in the frontend is we can give better status updates to users.

Perhaps I'm outnumbered on this one, but that does get rid of the timing guarantees of using a TaskSequence.

friendshipper/src-tauri/src/repo/operations/reset.rs
```diff
@@ -1,12 +1,177 @@
...
+pub struct ResetToCommitOp {
+    pub commit: String,
+    pub git_client: git::Git,
+}
+
+#[async_trait]
+impl Task for ResetToCommitOp {
+    async fn execute(&self) -> Result<(), CoreError> {
+        self.git_client
+            .run(&["reset", "--hard", &self.commit], Default::default())
```
rudoi, line 38: `--hard` will get rid of all local changes. is that the intent here?

cynful: using `--keep` 👍

friendshipper/src-tauri/src/repo/operations/reset.rs
```diff
@@ -1,12 +1,214 @@
...
+        engine: state.engine.clone(),
+        engine_path,
+    };
+    sequence.push(Box::new(download_dlls_op));
+
+    // update engine
+    let app_config = state.app_config.read().clone();
+    let uproject_path =
+        PathBuf::from(&app_config.repo_path).join(&state.repo_config.read().uproject_path);
+    let new_uproject: Option<UProject> = match UProject::load(&uproject_path) {
```
rdunnington, line 145: Since `sequence` is an async operation, this `new_project` will have the exact same contents as `old_project` because the repo hasn't been updated yet. I recommend putting all the logic that's in the handler here into `ResetToCommitOp::execute()` so all the operations happen as part of one "transaction" - this is how `PullOp` works right now.

rdunnington (changes requested):

friendshipper/src-tauri/src/repo/operations/reset.rs
cynful: this has been moved into the :execute() and out of the handler

PR #285, opened 2024-09-26
chore(friendshipper): stamp EpicInternal.txt sentinel for PII in crash uploads

rdunnington:
* Unreal looks for the file Engine/Restricted/NotForLicensees/Build/EpicInternal.txt when it starts up to determine if the build can contain PII such as username and commandline. This info is important for engineers when tracing the source and cause of crashes, so we'll stamp it when the engine or client gets downloaded.


rudoi (approved):
awesome, i like what you've done making sure this fits within the provider interface

PR #290, opened 2024-10-03
feat(friendshipper): increase http buffer size in git config

sylviacx:
7561492698


rudoi (dismissed):

chris-believer:
We're good with allocating 1GB of memory on every git push?

sylviacx:
> We're good with allocating 1GB of memory on every git push?

Yea that might be over allocating for now. I can drop it back down to 524288000 and if we see any more connection issues we can bump it back up.

chris-believer (approved):

PR #292, opened 2024-10-07
chore(friendshipper): prepare mac dmg builds

rudoi:


.github/workflows/tauri.yml
```diff
@@ -0,0 +1,83 @@
+name: "publish"
```
rudoi, line 1: this is a test to migrate us over to the "tauri-recommended" workflow.

cynful (approved):

PR #305, opened 2024-10-10
feat(birdie): hook up otel via repo-level config file

rudoi:
Also instrumented a bunch of stuff.

### DIFF
--- a/birdie/src-tauri/src/main.rs
+++ b/birdie/src-tauri/src/main.rs
@@ -3,6 +3,12 @@
 // Prevents additional console window on Windows in release, DO NOT REMOVE!!
 #![cfg_attr(not(debug_assertions), windows_subsystem = "windows")]
 
+use std::fs;
+use std::path::PathBuf;
+
+use birdie::types::config::BirdieRepoConfig;
+use ethos_core::clients::git::Git;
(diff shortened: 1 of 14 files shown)


birdie/src-tauri/src/main.rs
```diff
@@ -64,168 +70,242 @@ fn force_window_to_front(window: Window) {
...
+                    });
+
+                    let username = git_client.get_username().await?;
+                    logging::init(
+                        VERSION,
+                        APP_NAME,
+                        VERSION,
+                        Some(username),
+                        repo_config.otlp_endpoint.clone(),
+                        repo_config.otlp_headers.clone(),
```
rdunnington, line 127: If `otlp_endpoint` or `otlp_headers` are empty strings, maybe we return an error before getting here?

rdunnington (approved):

birdie/src-tauri/src/main.rs
rudoi: the internal logging init function does check for empty, does that fit the bill?

rudoi: just pushed up

rdunnington (approved):

PR #314, opened 2024-10-11
feat(birdie): on view switch, remember file tree state, + fix compile…

sylviacx:
draft while I implement the default view save


birdie/src/lib/components/files/TreeNode.svelte
```diff
@@ -56,12 +56,6 @@
 		loading = false;
 	};
 
-	onMount(async () => {
-		loading = true;
-		await getChildren();
-		loading = false;
-	});
-
```
sylviacx, line 64: Turns out this on mount was overwriting child nodes and losing all their children, resulting in only the top level node being applied properly

cynful (approved):
👍 ✨

rudoi (approved):
nice! great work 🍕

PR #315, opened 2024-10-12
chore(friendshipper): poll for pr mergeability in a separate thread

rudoi:
Also, disable deep link registration for mac.

### DIFF
--- a/friendshipper/src-tauri/src/repo/operations/gh/submit.rs
+++ b/friendshipper/src-tauri/src/repo/operations/gh/submit.rs
@@ -79,7 +79,7 @@ impl Task for GitHubSubmitOp {
             repo = status.repo_name.clone();
         }
 
-        let mut pr = octocrab
+        let pr = octocrab
             .pulls(owner.clone(), repo.clone())
             .create(
                 format!("{} {}", SUBMIT_PREFIX, truncated_message),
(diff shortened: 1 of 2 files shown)


friendshipper/src-tauri/src/repo/operations/gh/submit.rs
```diff
@@ -89,16 +89,38 @@ impl Task for GitHubSubmitOp {
...
+                        if let Err(e) = client_clone.enqueue_pull_request(id).await {
+                            warn!("Failed to enqueue pull request: {:?}", e);
+                        }
+                    } else {
+                        warn!("Failed to get pull request ID");
+                    }
+                }
+                Err(e) => warn!("Failed to poll for mergeable state: {:?}", e),
+            }
+        });
```
rudoi, line 123: The potential downside here is if queuing the PR errors, the frontend might not be aware of it. I'd like any reviewer to chime in on whether that risk is worth the 10-30s of savings on submission time.

cynful: I dont think the saved time is enough to offset a bad ui experience?
maybe if users were used to checking on their submissions via github, this would be fine.

rudoi: the bad ui experience is if and only if putting the PR into the queue errors out. the frontend still watches the PR and updates its state accordingly. the difference is that we were blocking the submission modal from closing until the PR gets queued. 

If the background thread errors out, users will see that their PR is open but not in the queue. The fix in this case would be to click the PR link in the UI and manually put the PR into the queue. In our data, this error has occurred 0 times in the last 60 days.

cynful: this seems fine then

cynful (approved):

PR #328, opened 2024-10-16
fix(birdie): breadcrumb navigation selects files

sylviacx:


sylviacx:
Without breadcrumb affecting selectedFile, context switching from default -> file tree doesn't work because that switch relies on selectedFile being not null

rudoi (approved):

rudoi (approved):

PR #335, opened 2024-10-21
feat(friendshipper): display junit output for builds

sylviacx:
#7514531708
#7514542393


core/src/utils/junit.rs
```diff
@@ -2,10 +2,34 @@ use serde::{Deserialize, Serialize};
 
 use crate::types::errors::CoreError;
 
+pub const TEST_JUNIT_XML: &str = r#"<?xml version="1.0" encoding="UTF-8"?>
```
rudoi, line 5: i think this can go away :3

rudoi (approved):

PR #352, opened 2024-10-30
fix(friendshipper): don't store empty strings for tokens

rudoi:


friendshipper/src/routes/+layout.svelte
```diff
@@ -277,8 +281,12 @@
 			try {
 				await tryOktaRefresh();
 			} catch (_) {
-				await handleOktaLogout();
-				await handleOktaLogin();
+				try {
```
cynful, line 284: oh the nested try catch

so, if refresh fails, then do a logout and login, nice

cynful (approved):

chris-believer (approved):

PR #361, opened 2024-11-07
fix(core): redact github_pat from tracing/debug

cynful:


rdunnington (approved):

core/src/types/config.rs
```diff
@@ -42,6 +43,27 @@ pub struct OktaConfig {
...
+    fn from(s: String) -> Self {
+        RedactedString(s)
+    }
+}
+
+impl std::fmt::Display for RedactedString {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", &self.0)
+    }
+}
```
rudoi, line 65: Thinking about it, what do you think about implementing `ToString` directly here? https://doc.rust-lang.org/std/string/trait.ToString.html

We may need to suppress a linter error that says "you should just implement `Display`". The rationale here is if you implement `Display` for this, the token could still get logged via a `"{}"` format string.

rudoi (approved):

PR #365, opened 2024-11-11
fix(core): filter out existing locks, fix changeset filter logic

rudoi:


core/src/clients/github.rs
```diff
@@ -140,7 +140,6 @@ impl GraphQLClient {
         .await
         {
             Ok(res) => {
-                warn!("get_pull_requests: {:?}", res);
```
rudoi, line 143: this is spamming big time

chris-believer (approved):

PR #366, opened 2024-11-12
fix(birdie): revert and lock selected should refer to selectedFiles list

sylviacx:


birdie/src/routes/+page.svelte
```diff
@@ -575,7 +575,7 @@
 		asyncModalText = 'Locking Files';
 
 		try {
-			const selectedPaths = $selectedExplorerFiles.map((file) => file.path);
+			const selectedPaths = $selectedFiles.map((file) => file.path);
```
sylviacx, line 578: big oops, the revert/lock selected buttons were referring to the files selected in the explorer instead of in the modified files card

PR #385, opened 2024-12-16
fix(friendshipper): fix initial repo clone from modal

rudoi:


friendshipper/src/lib/components/oobe/WelcomeModal.svelte
```diff
@@ -151,26 +150,23 @@
...
 		errorMessage = '';
 		try {
+			cloning = true;
 			await handleUpdateAppConfig();
 
-			// only clone if this is our first time through
-			if (!currentConfig.repoPath) {
-				cloning = true;
-				await cloneRepo({ url: repoUrl, path: cloneLocation });
+			await cloneRepo({ url: repoUrl, path: cloneLocation });
```
chris-believer, line 156: Is pulling the repoPath check safe if `cloneLocation` already exists?

rudoi: yep! it's an early out in the backend

chris-believer (approved):

PR #386, opened 2024-12-17
chore(birdie): show commit history for folders

rudoi:
Also:
- Disabled some file metadata stuff that we aren't really using yet
- Rebalanced the UI a bit to allow more space for commits


cynful (approved):
:shipit:

PR #395, opened 2025-01-09
chore(core): kill process conflicting on port

rudoi:


core/src/utils/process.rs
```diff
@@ -67,6 +68,41 @@ pub fn check_for_process(name: &str) -> Result<()> {
...
 
+            match listeners::get_processes_by_port(port) {
+                Ok(listeners) => {
+                    for listener in listeners {
+                        if listener.pid != my_pid {
+                            if let Some(process) = system.process(Pid::from_u32(listener.pid)) {
+                                warn!(
+                                    "Found existing process {} on port {} but couldn't reach its API. Attempting to kill it.",
+                                    process.name(),
+                                    listener.pid
```
rdunnington, line 79: Is this the port or process ID?

core/src/utils/process.rs
```diff
@@ -67,6 +68,41 @@ pub fn check_for_process(name: &str) -> Result<()> {
                 }
             }
 
+            match listeners::get_processes_by_port(port) {
```
rdunnington, line 71: Maybe I'm blind, but I can't see where `port` is actually used

core/src/utils/process.rs
rudoi: ah, good call, this should be port

core/src/utils/process.rs
rudoi: you mean where it comes from? it's a parameter to the function

rudoi: `listeners` is a crate which has `get_processes_by_port` as a static helper

rdunnington: nvm I'm blind

rdunnington (approved):

chris-believer (approved):

PR #396, opened 2025-01-09
fix(core): move repo refetch and commit-graph rewrite to dangerzone

chris-believer:
This should reduce the startup time hit, but still allow access if necessary during troubleshooting.


rudoi (approved):
looks great!

PR #403, opened 2025-01-31
fix(core): make sure file meta updates in modified files card

rudoi:
When taking on updated to modified files in the ModifiedFilesCard component, we were ignoring any updates to the metadata in the files, and only removing files no longer in the modified files array.

### DIFF
--- a/core/src/longtail.rs
+++ b/core/src/longtail.rs
@@ -140,10 +140,7 @@ impl Longtail {
         // Check the path found above, and if all else fails try to find it in $PATH
         match which_in(exe_name.clone(), exe_path, env::current_dir().unwrap()) {
             Ok(path) => Some(path),
-            Err(_) => match which(exe_name) {
-                Ok(path) => Some(path),
-                Err(_) => None,
-            },
+            Err(_) => which(exe_name).ok(),
         }
     }
 
--- a/core/ui/src/lib/components/repo/ModifiedFilesCard.svelte
+++ b/core/ui/src/lib/components/repo/ModifiedFilesCard.svelte
@@ -120,10 +120,11 @@
 	const cleanUpChangeSets = async () => {
 		ensureDefaultChangeset();
 		for (let i = 0; i < changeSets.length; i += 1) {
-			// Keep only files that exist in modifiedFiles
-			changeSets[i].files = changeSets[i].files.filter((file) =>
-				modifiedFiles.some((mf) => mf.path === file.path)
-			);
+			// Keep only files that exist in modifiedFiles, while updating the file state to match the entry in modifiedFiles
+			changeSets[i].files = changeSets[i].files.reduce<ModifiedFile[]>((acc, file) => {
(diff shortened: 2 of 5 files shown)


core/src/longtail.rs
```diff
@@ -140,10 +140,7 @@ impl Longtail {
         // Check the path found above, and if all else fails try to find it in $PATH
         match which_in(exe_name.clone(), exe_path, env::current_dir().unwrap()) {
             Ok(path) => Some(path),
-            Err(_) => match which(exe_name) {
-                Ok(path) => Some(path),
-                Err(_) => None,
-            },
+            Err(_) => which(exe_name).ok(),
```
rudoi, line 143: clippy `manual implementation of .ok()` error, same with below

core/ui/src/lib/components/repo/ModifiedFilesCard.svelte
```diff
@@ -120,10 +120,11 @@
...
 		for (let i = 0; i < changeSets.length; i += 1) {
-			// Keep only files that exist in modifiedFiles
-			changeSets[i].files = changeSets[i].files.filter((file) =>
-				modifiedFiles.some((mf) => mf.path === file.path)
-			);
+			// Keep only files that exist in modifiedFiles, while updating the file state to match the entry in modifiedFiles
+			changeSets[i].files = changeSets[i].files.reduce<ModifiedFile[]>((acc, file) => {
+				const updatedFile = modifiedFiles.find((mf) => mf.path === file.path);
+				return updatedFile ? [...acc, updatedFile] : acc;
+			}, []);
```
rudoi, line 127: `reduce` syntax can be confusing to look at, i have to read the docs every time. basically it uses an accumulator `acc`, and passes that into each run through the loop. `acc` is a single value, in this case a new array of files. if we find a file in `modifiedFiles`, it's allowed to stay in the changeset, so we push it into the accumulator. if it's not in modifiedFiles, it's not allowed to stay in the changeSet, so we do nothing to `acc`. 

tldr, this uses `reduce` to both:
1. filter out anything in the changeSet that isn't in modifiedFiles anymore, and
2. update file info (like lock status) for each file that DOES exist in modifiedFiles

rudoi: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

friendshipper/src/lib/components/preferences/PreferencesModal.svelte
```diff
@@ -221,6 +228,7 @@
 			showProgressModal = true;
 			progressModalTitle = 'Resetting repo...';
 			await resetRepo();
+			$repoStatus = await getRepoStatus();
```
rudoi, line 231: i noticed resetting the repo to main doesn't also update the status, so snuck that in here

rdunnington (approved):

PR #406, opened 2025-02-03
fix(friendshipper): make sure hour converts to locale string in playt…

rudoi:
…est modal

### DIFF
--- a/friendshipper/src/lib/components/playtests/PlaytestModal.svelte
+++ b/friendshipper/src/lib/components/playtests/PlaytestModal.svelte
@@ -206,9 +206,9 @@
 
 	const getPlaytestTime = (item: Nullable<Playtest>): string => {
 		const date = item != null ? new Date(item.spec.startTime) : new Date();
-		return `${date.getHours()}:${date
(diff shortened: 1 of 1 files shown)


friendshipper/src/lib/components/playtests/PlaytestModal.svelte
```diff
@@ -206,9 +206,9 @@
 
 	const getPlaytestTime = (item: Nullable<Playtest>): string => {
 		const date = item != null ? new Date(item.spec.startTime) : new Date();
-		return `${date.getHours()}:${date
-			.getMinutes()
-			.toLocaleString('en-US', { minimumIntegerDigits: 2 })}:00`;
+		const hours = date.getHours().toLocaleString('en-US', { minimumIntegerDigits: 2 });
```
rudoi, line 209: i believe the bug here was that the timer selection is always expecting 2 digits for hours and we weren't providing that, meaning this would always be broken from 00-09 oclock haha.

rdunnington (approved):

PR #412, opened 2025-02-07
chore(friendshipper): drop unreal requests while it's in a slow task

rdunnington:
* Because Unreal can go into a blocking "slow task mode" on its' main thread and not ever respond to requests we issue, expose a new handler that Unreal can ping to say when it's going into/out of a slow task. When Unreal is in a slow task, we drop all existing requests to Unreal and don't allow new ones.

### DIFF
--- a/friendshipper/src-tauri/src/engine/unreal/ofpa.rs
+++ b/friendshipper/src-tauri/src/engine/unreal/ofpa.rs
@@ -185,6 +185,8 @@ impl OFPANameCache {
                     filenames: paths_to_request.clone(),
                 };
 
+                // See UnrealEngineProvider::send_status_update() implementation for an explanation of this approach to
+                // manually polling requests sent to Unreal.
                 let client = reqwest::Client::new();
                 let res_or_err_future = client
                     .post("http://localhost:8091/friendshipper-ue/ofpa/friendlynames".to_string())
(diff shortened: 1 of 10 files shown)


friendshipper/src-tauri/src/engine/unreal/provider.rs
```diff
@@ -79,13 +82,30 @@ impl EngineProvider for UnrealEngineProvider {
...
             let client = reqwest::Client::new();
-            _ = client
+            let future = client
                 .post("http://localhost:8091/friendshipper-ue/status/update".to_string())
                 .json(status)
-                .send()
-                .await;
+                .send();
+
+            tokio::pin!(future);
```
rdunnington, line 96: Extremely helpful forum post: https://users.rust-lang.org/t/how-to-check-if-a-future-is-immediately-ready/86401

friendshipper/src-tauri/src/engine/unreal/ofpa.rs
```diff
@@ -175,47 +176,95 @@ impl OFPANameCache {
...
-                    } else {
-                        match res.json::<OFPAFriendlyNamesResponse>().await {
-                            Ok(data) => {
-                                let mut cache = provider.ofpa_cache.write();
-                                for item in data.names {
-                                    if item.error.is_empty() {
-                                        cache.add_name(&item.file_path, &item.asset_name, now);
+                    .send();
+
+                let res_or_err = 'block: {
```
rudoi, line 194: I think this could use some fat comments, wdyt?

friendshipper/src-tauri/src/engine/unreal/provider.rs
rudoi: I'd even consider linking this in a comment directly

rudoi (approved):
yay futures magic

rudoi (approved):

PR #415, opened 2025-02-08
chore(core): fix linux dependency break

rdunnington:
* graphql_client seems to not lock their dependency on zbus, which recently had a breaking change in the new 5.5.0 version. This change ensures we lock to versions less than 5.5.0 to avoid a build break on linux.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2167,6 +2167,7 @@ dependencies = [
  "walkdir",
  "which",
  "windows 0.56.0",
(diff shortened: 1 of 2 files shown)


rudoi (approved):
god's work

rdunnington:
Turns out we can't really resolve this on our end because tauri pins to a version of `rust-notify` that has this break. I opened a PR here to potentially resolve the issue. https://github.com/hoodie/notify-rust/pull/238

PR #424, opened 2025-02-13
feat(friendshipper): add ability to add new projects to friendshipper through preferences

sylviacx:


friendshipper/src/lib/components/preferences/PreferencesModal.svelte
```diff
@@ -92,13 +93,56 @@
...
+		configuringNewRepo = true;
+		localAppConfig.projects['new-project'] = {
+			repoPath: '',
+			repoUrl: ''
+		};
+		localAppConfig.selectedArtifactProject = 'new-project';
+	};
+
+	const onRepoUrlInput = (e: Event) => {
+		const input = (e.target as HTMLInputElement).value;
```
sylviacx, line 106: @rudoi I removed your early return if we are not configuring a new repo because that prevented the artifact name from updating with the repo url

rudoi: sounds good! 

Sent from my iPhone.

friendshipper/src-tauri/src/config/router.rs
```diff
@@ -108,12 +108,6 @@ where
         .into());
     }
 
-    if current_config.selected_artifact_project.is_some() {
-        payload
-            .selected_artifact_project
-            .clone_from(&current_config.selected_artifact_project)
-    }
-
```
sylviacx, line 116: Needed to remove this because this prevented the new selected artifact project from being written to the app config

rudoi (approved):

friendshipper/src-tauri/src/config/router.rs
```diff
@@ -215,11 +210,14 @@ where
 
             payload.repo_path = full_repo_path.to_str().unwrap().to_string();
 
-            // call the DLL download handler
-            let _ = download_dlls_handler(State(state.clone())).await?;
+            // TODO(sylviacx): guard this because we don't have a way to test cloning a new project AND setting up dlls and engine stuff
```
sylviacx, line 213: We can remove this guard whenever we get a new project that we can actually test dll and engine stuff with, but for now while we test with normal git repos, we will always error out on these calls

friendshipper/src/lib/components/preferences/PreferencesModal.svelte
```diff
@@ -126,14 +169,21 @@
...
+		localAppConfig.repoUrl =
+			localAppConfig.projects[localAppConfig.selectedArtifactProject].repoUrl;
+
 		const internal = async () => {
 			try {
 				const accessToken = $oktaAuth?.getAccessToken();
 				if (accessToken) {
-					await updateAppConfig(localAppConfig, accessToken);
+					progressModalTitle = 'Saving preferences...';
+					await updateAppConfig(localAppConfig, accessToken, true);
```
sylviacx, line 183: updateAppConfig has logic where if a git repo is not set at the repoPath, it will try and clone it so we don't need any extra logic outside of that (for now)

friendshipper/src-tauri/src/config/router.rs
rudoi: it sounds like there's some work here to make the repos work "without DLLs"

rudoi (approved):

PR #426, opened 2025-02-20
chore(core): upgrade tauri apps to v2

rudoi:


cynful (approved):
🎉

core/Cargo.toml
```diff
@@ -29,7 +29,7 @@ serde_json = { workspace = true }
 serde_with = { workspace = true }
 serde_yaml = { workspace = true }
 sysinfo = { workspace = true }
-tauri = { workspace = true }
+tauri = { version = "2.2.5" }
```
chris-believer, line 32: This is the only tauri that's pinned to `2.2.5`, it looks like the others are both `2`. The root `Cargo.toml` still has 1.5 listed, but I don't think that matters anymore. Worth pulling back into the workspace as a followup?

chris-believer (approved):

core/Cargo.toml
rudoi: yeah i'll clean that up 🖖🏻

PR #430, opened 2025-02-21
fix(core): use correct updater API

rudoi:


core/src/tauri/command.rs
```diff
@@ -66,21 +63,7 @@ pub async fn restart(state: tauri::State<'_, State>) -> Result<(), TauriError> {
...
-                tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
-
-                std::process::exit(0);
-            }
-            Err(e) => error!("Error restarting: {}", e),
-        }
-    }
-
-    Ok(())
+    tauri::process::restart(&Env::default());
```
rudoi, line 66: switching over to our own restart function lets us give the backend a few seconds to shut down

PR #435, opened 2025-02-21
feat(friendshipper): add ability to cancel client sync

rudoi:


friendshipper/src-tauri/src/builds/router.rs
```diff
@@ -228,19 +230,49 @@ where
...
         Ok(_) => {
-            state.longtail.get_archive(
-                &local_path,
-                None,
-                &archive_urls,
-                tx,
-                aws_client.get_credentials().await,
-            )?;
+            let (cancel_tx, mut cancel_rx) = oneshot::channel();
+            state.cancel_tx.write().await.replace(cancel_tx);
```
rdunnington, line 237: I don't really understand the purpose of this line, could use a comment

rdunnington: ah I think I see, basically this is storing the cancel channel into the app state

rdunnington (dismissed):

rdunnington (approved):

PR #440, opened 2025-02-24
feat(friendshipper): allow opening files in editor

rudoi:


rdunnington (commented):
using that sweet tauri 2.0 custom context menu feature

PR #451, opened 2025-03-05
feat(friendshipper): quick submit directly pushes to content-main whe…

sylviacx:
…n submitting on it


friendshipper/src-tauri/src/repo/operations/push.rs
```diff
@@ -0,0 +1,183 @@
+use crate::engine::EngineProvider;
```
sylviacx, line 1: This push handler is mostly copied over from the birdie push handler

PR #464, opened 2025-03-19
chore(friendshipper): use shared client cache directory

rudoi:


chris-believer (approved):

friendshipper/src/lib/components/preferences/PreferencesModal.svelte
```diff
@@ -421,18 +424,30 @@
...
+					/>
+				</div>
 			</div>
 			<Tooltip class="text-sm items-center" placement="bottom">
 				Group downloaded builds by playtest. This will allow you to keep multiple playtests synced
 				at once. However, your initial sync of each playtest <span class="font-bold"
 					>will take longer</span
-				>. This option also uses significantly more disk space.
+				>. This option also uses significantly more disk space. The max cache size controls when
+				older cached files will be deleted.
```
chris-believer, line 450: Worth calling out that old clients will never be deleted automatically, and will need to be cleaned up manually?

chris-believer (dismissed):

chris-believer (approved):

PR #466, opened 2025-03-20
perf(friendshipper): File History performance

blvr-kbrooks:
- File History now gets all previous revision history async instead of serially


rudoi (approved):
thanks!

rudoi (approved):

PR #468, opened 2025-03-28
feat(friendshipper): add target branches list to repo config, submit …

sylviacx:
…fn determined by branch config


friendshipper/src-tauri/src/repo/operations/gh/auto_merge_submit.rs
```diff
@@ -381,12 +376,12 @@ where
...
             let gh_op = GitHubSubmitOp {
                 head_branch: worktree_branch.clone(),
-                base_branch: base_branch.to_string(),
+                base_branch: target_branch.to_string(),
                 token: self.token.clone(),
                 commit_message: self.commit_message.clone(),
                 repo_status: self.repo_status.clone(),
                 client: self.github_client.clone(),
-                enable_auto_merge: false,
+                enable_auto_merge: true,
```
sylviacx, line 384: was false before because we inverted the fns for testing purposes, this should be true because we want to auto merge

friendshipper/src-tauri/src/repo/operations/gh/submit.rs
```diff
@@ -606,12 +600,12 @@ where
...
                     let gh_op = GitHubSubmitOp {
                         head_branch: worktree_branch.clone(),
-                        base_branch: base_branch.clone(),
+                        base_branch: target_branch.clone(),
                         token: self.token.clone(),
                         commit_message: self.commit_message.clone(),
                         repo_status: self.repo_status.clone(),
                         client: self.github_client.clone(),
-                        enable_auto_merge: true,
+                        enable_auto_merge: false,
```
sylviacx, line 608: was true before because we inverted the fns for testing purposes, this should be false because this is the original quick submit fn and we want to use merge queue

rudoi (approved):
looks great, thanks for doing this!

PR #474, opened 2025-04-04
Build deeplink updates

rdunnington:


friendshipper/src/lib/components/workflows/WorkflowLogsModal.svelte
```diff
@@ -124,7 +123,17 @@
 
 	const copyLogToClipboard = async () => {
 		try {
-			await writeText(rawLogs);
+			await navigator.clipboard.writeText(rawLogs);
```
rudoi, line 126: oh ok so this actually switches us OFF the tauri clipboard plugin? interesting, i wonder what the downside is

rudoi (approved):

PR #476, opened 2025-04-10
feat(friendshipper): reset engine in danger zone

rdunnington:
* There have been several cases where people have reported broken engines after switching to a branch with a different engine and updating. This adds a mitigation where users can have a convenient button that will delete all prebuilt engine binaries and data, and redownload it.


rudoi (approved):
good idea, and thanks!

PR #486, opened 2025-04-21
ci(core): temporarily remove signing thumbprint

rudoi:
CI is currently acting up because of some inconsistencies in the cert chain.


chris-believer (commented):
Approve! :smiley:

PR #510, opened 2025-05-30
feat: allow optional static credentials during build

chris-believer:
This was supposed to be straightforward, but a ton of clippy fixes came in as well. The only other things to note are an instrument on the longtail `log_message` to get them into otel and a number of null checks in typescript where it was throwing uncaught exceptions when stores weren't initialized. Those may not be needed in practice, but seemed to be better to have anyway?


rudoi (approved):
lgtm!

PR #517, opened 2025-07-23
fix(friendshipper): hardcode desired git version for winget

rudoi:
Also fixes an issue where non-repo users can't see playtests.


chris-believer (approved):
Looks great! Pinning to a known version is better here. Has anyone looked at why 2.50 isn't working?

itscoho (approved):

PR #520, opened 2025-07-24
fix(friendshipper): Update default player count for playtest from 1 to 4 and fixing some precommit failures

itscoho:


birdie/src-tauri/src/repo/push.rs
```diff
@@ -16,14 +15,6 @@ use crate::state::AppState;
 
 use super::StatusOp;
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
```
itscoho, line 19: unused so removing

friendshipper/src-tauri/src/system/unreal.rs
```diff
@@ -137,18 +137,17 @@ pub fn check_unreal_file_association() -> anyhow::Result<(bool, Vec<String>)> {
         {
             info!("Checking .uproject file association");
             let uproject_key = hkcr.open_subkey(".uproject");
-            if uproject_key.is_err() {
```
itscoho, line 140: These 2 changes were precommit detected problems with the previous "unwrap" approach

PR #531, opened 2025-08-14
Create UI to show what changes are in the primary branch but have yet to be merged to content branch

itscoho:


itscoho:
in addition, this was a chance for me to do something a bit more complex with how our app works so that's always a plus ;)

itscoho:
image

chris-believer (dismissed):
Looks good to me!

Three general questions, but approved if none of these are a concern:
* What is the behavior when `validate_configured_branches` errors? Since it's in `initialize_server`, is it too early to show up as an error message in the UI?
* Does the playtest-only config have any trouble starting or building the home page? (I don't think `ContributorLayout` is used in that config, but not certain.)
* Any concern with the speed of the `git cherry` operation? Any significant delay in it returning results during testing?

itscoho:
Thanks Chris, I'll double check those and get back to you.  FWIW, the git cherry seems pretty snappy (at least from my testing :) )

itscoho:
@chris-believer great catch! needed a bit more plumbing to pass along the error, but it's good now.  In addition, validated that the playtest mode works w/o a hitch

chris-believer (approved):

PR #540, opened 2025-08-22
fix(friendshipper): allow for syncs with untracked files with additio…

itscoho:
…nal conflict checks

### DIFF
--- a/core/src/clients/git.rs
+++ b/core/src/clients/git.rs
@@ -502,21 +502,33 @@ impl Git {
     ) -> anyhow::Result<()> {
         self.wait_for_lock().await;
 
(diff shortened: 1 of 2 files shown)


itscoho:
A few new details:
1. We check if there's conflicts with untracked files on pull
2. We handle if you restore a snapshot and it has conflicts that are different from your local untracked version, save a copy and set it aside (while still restoring the snapshot) and then error to let you know 👍

itscoho:
I initially went down a path where we stashed the untracked files in restoring snapshots but it felt bad doing snapshots while restoring them when something much simpler like this could work.

PR #541, opened 2025-08-26
fix(friendshipper): fix quicksubmit taking you back to target branch …

itscoho:
…if there's no merge queue


lmcdowell (approved):
Other than those annotation errors. LG

PR #542, opened 2025-08-26
fix(friendshipper): allow for dynamic ports on launch and .port file …

itscoho:
…for ue plugin

### DIFF
--- a/friendshipper/src-tauri/tests/common/mod.rs
+++ b/friendshipper/src-tauri/tests/common/mod.rs
@@ -89,7 +89,7 @@ where
     where
         T: EngineProvider,
     {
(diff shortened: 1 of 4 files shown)


friendshipper/src-tauri/tests/common/mod.rs
```diff
@@ -89,7 +89,7 @@ where
     where
         T: EngineProvider,
     {
-        let app = friendshipper::router(&self.state.log_path)
+        let app = friendshipper::router(&self.state.log_path, 8585)
```
lmcdowell, line 92: What's this hardcoded number?

chris-believer (approved):

PR #545, opened 2025-08-28
fix(friendshipper): fix snapshot management when restoring after a sync

itscoho:


lmcdowell (dismissed):
no notes. LGTM

core/src/clients/git.rs
```diff
@@ -529,6 +530,26 @@ impl Git {
...
+                    Opts::default(),
+                )
+                .await
+                .ok();
+            }
+        }
+
         let mut apply_args = vec!["cherry-pick", "-n", "-m1"];
 
         // if we're keeping ours, use the theirs strategy because this is a stash pop
```
itscoho, line 555: @rudoi I found this solution to the issue with syncing while having local changes (Cherry-pick doesn't like it when you reapply the snapshot after the sync, however it is separate from the conflict strategy we currently have here. Thoughts?  I'm not sure if people have used the conflict strategy so I figured it was worth discussion.

rudoi: i don't think anyone uses that setting at all so, if this fixes the issue im good with it

itscoho:
OK 2 changes: 

1) this new approach to restoring snapshots makes conflict strategy resolution obsolete (but we still keep errors for syncing with conflicts as that's what most people use today anyway)
2) removed the auto branch switch for quicksubmits on non-mergequeue branches since it's problematic if the editor running and we need those branches to allow for submission without a pull. Would rather we keep the functionality consistent rather than trying to be fancy about it.

friendshipper/src-tauri/src/repo/operations/gh/submit.rs
```diff
@@ -684,16 +682,6 @@ where
 
             // unlock all files submitted
             lock_op.execute().await?;
-
```
itscoho, line 687: I added this for the last FS version, but it's got some def. issues so backing it out for now

PR #549, opened 2025-09-04
fix(friendshipper): remove base f11r branch push for quicksubmit sinc…

itscoho:
…e it's not used in the pr


itscoho:
as per our discussion, we only need the worktree branch for our workflows so we should stop pushing the base f11r branch as part of the quick submit flow

rudoi (dismissed):
seems fine! i think we initially added this just so people's work would be saved remotely, but now that we use the worktree always seems legit

PR #563, opened 2025-10-16
fix(friendshipper): adding git gc at launch to keep it under control …

itscoho:
…a bit more directly


itscoho:
closing this out.  The maintenance should be doing the job here, but going to add a check to object count on the diagnostics panel to give us more insights here.

PR #573, opened 2026-02-23
feat(friendshipper): add a checkbox (default true) to sync on quicksu…

itscoho:
…bmit for content branch


itscoho:
Going to let this sit as a draft for a bit since this feels less clear. Adding this would add the requirement to close the editor to quicksubmit.  I'd rather us not add additional constraints and instead be clearer about what files you have locally wrt locks.  ie: we should just make it obvious WHEN you have to sync (i want to lock or modify an out-of-date asset for example).  if we manage that correctly, I think this makes this feature less needed (though it's still a little weird to be on a branch intended for merge queue when there is none). Big <3 to @chris-believer for good discussion here

PR #593, opened 2026-05-11
feat(friendshipper): zip/import local changes with pre-import snapshot

itscoho:
Bundles selected files into a portable zip (with manifest) and imports zips back into the working tree with a preview. Auto-snapshots any local paths the import would clobber. Options can be found in a drop-down next to the snapshot button on the submit page.

fix(core): preserve full snapshot message when it contains "snapshot" 
fix(core): stage deletions via git rm --cached so save_snapshot works
           for selections that include removed files

### DIFF
--- a/core/src/clients/git.rs
+++ b/core/src/clients/git.rs
@@ -358,7 +358,13 @@ impl Git {
                 }
 
                 let stash_index = parts[0].trim();
-                let message = parts[1].split(SNAPSHOT_PREFIX).collect::<Vec<_>>()[1].trim();
+                // Split only at the first occurrence so messages that themselves contain
+                // the word "snapshot" (e.g. "Auto-snapshot before importing X.zip") are
+                // preserved in full instead of being truncated at the embedded match.
+                let message = match parts[1].split_once(SNAPSHOT_PREFIX) {
+                    Some((_, m)) => m.trim(),
+                    None => return None,
+                };
                 let commit = parts[2].trim();
                 let date = parts[3].trim();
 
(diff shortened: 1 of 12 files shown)


core/src/clients/git.rs
```diff
@@ -399,23 +405,52 @@ impl Git {
...
-                ],
-                Opts::default(),
-            )
-            .await?;
+            if !deleted_paths.is_empty() {
+                let mut rm_temp = NamedTempFile::new()?;
+                for path in &deleted_paths {
+                    writeln!(rm_temp, "{path}")?;
+                }
+                rm_temp.flush()?;
```
chris-believer, line 438: nit - I think this should be using [into_temp_path](https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html#method.into_temp_path), because we're passing it to git while we're still holding it open. It's obviously been working, though... :D

core/src/clients/git.rs
```diff
@@ -399,23 +405,52 @@ impl Git {
...
+                // `--cached` keeps git from trying to remove files from the working tree
+                // (they're already gone); `--ignore-unmatch` keeps the command from
+                // failing if a path happens to be missing from the index.
+                self.run(
+                    &[
+                        "rm",
+                        "--cached",
+                        "--ignore-unmatch",
+                        "--pathspec-from-file",
+                        rm_temp.path().to_str().unwrap(),
```
chris-believer, line 448: Another nit, obviously not a problem so far - if the path is non-UTF-8 this unwrap would blindly fail. Seems like it should map_err or at least be a expect?

rudoi (approved):

chris-believer (commented):
minor things, lgtm!

PR #594, opened 2026-05-11
Coho/fix snapshot bugs

itscoho:


itscoho:
This PR fixes a couple bugs where untracked files weren't captured properly in a number of snapshot situations and selective file operations like reverts from a list of your files also would capture the wrong amount of files in the snapshot.  In addition, this provides a preview of what files can be restored from a snapshot and allows you to choose to stomp your local file state from what is in the snapshot if you desire.


Finally, I removed some dead code around the _keep_index parameter as not needed anymore since we now build a stash-shaped 2-parent commit against a temp index (GIT_INDEX_FILE), so narrow snapshots capture exactly what was requested, untracked files are included, and the user's real index is never touched. git cherry-pick -m1 still restores it.

core/src/clients/git.rs
```diff
@@ -386,30 +381,98 @@ impl Git {
...
+                &stash_message,
+                &snapshot_commit,
+            ],
+            Opts::default(),
+        )
+        .await?;
+
+        let snapshots = self.list_snapshots().await?;
+        let first = snapshots.first().cloned();
+        assert!(first.is_some(), "Failed to get snapshot");
```
chris-believer: This assert feels bad. If it ever fires the async task panic probably isn't going to be handled well. Should this be an anyhow::ensure! instead?

core/src/clients/git.rs
```diff
@@ -545,7 +610,25 @@ impl Git {
...
         currently_modified_files: Vec<File>,
         prefer_snapshot_versions: bool,
+        paths_filter: Option<Vec<String>>,
     ) -> anyhow::Result<()> {
+        // Selective restore takes a different, simpler path: we extract each
+        // selected file straight out of the snapshot's tree (or delete it,
+        // for paths the snapshot recorded as deleted) without touching the
+        // user's real index. The selective fn uses `prefer_snapshot_versions`
+        // as the "overwrite local changes" toggle — when false, it refuses
+        // to touch paths that have uncommitted local edits.
```
chris-believer: I'm having a bit of a hard time wrapping my head around the differences between the selective restore vs. the non-selective restore. 

It seems like the cleaner code in the new selective restore path. Wouldn't we achieve the same outcome if we build the paths_filter from all of the snapshot commits changes and pass that to restore_snapshot_selective if paths_filter==None? Then we only have one git mechanism for restoring a snapshot, regardless of it being selective or not.

core/src/clients/git.rs
```diff
@@ -713,6 +799,117 @@ impl Git {
...
+            }
+        }
+
+        // Apply deletions directly — no git invocation needed, and `fs::
+        // remove_file` doesn't care whether the file was LFS-tracked.
+        for path in &to_delete {
+            let abs = self.repo_path.join(path);
+            if abs.exists() {
+                std::fs::remove_file(&abs).map_err(|e| {
+                    anyhow!("failed to remove {} during selective restore: {}", path, e)
```
chris-believer: If we fail here (or anywhere else, really) we'll stop processing the snapshot restore. Would we want to collect the errors and try to continue, or put the changes back to a more consistent state? Might not be worth the trouble, but I could see the error toast being ignored pretty easily. 🤔

chris-believer (approved):
A few thoughts, but I think the change looks great. The only other thing that comes to mind would be adding tests to show the previous failure states, but that may end up being tough?

core/src/clients/git.rs
itscoho: Great idea, I'll add that :)

core/src/clients/git.rs
itscoho: agreed, fixing

core/src/clients/git.rs
itscoho: hmm I'll take a whack and unifying things 🤔 at the very least to keep behavior but use this new approach across the board

core/src/clients/git.rs
itscoho: Ok should be sorted now

core/src/clients/git.rs
itscoho: Ok should be better now, agreed, that's a better flow

chris-believer (approved):
lgtm!

PR #596, opened 2026-05-11
Coho/gate conflicts from quicksubmit

itscoho:


itscoho:
Need to think on this one a bit more.  pulling back to draft

core/src/clients/git.rs
```diff
@@ -2046,6 +2046,25 @@ mod tests {
             .await
             .expect("save_snapshot_all");
 
+        // Cross the filesystem mtime tick before restoring. This test is flaky
+        // on Linux CI (tmpfs, ~1s mtime resolution): the snapshot stamps a temp
```
chris-believer, line 2050: This is wrong, but probably doesn't matter? Tmpfs has nanosecond resolution, and later on this says NTFS has coarser resolution. 🤷 

If this explanation is masking a real bug but it's Linux only, I can dig but doesn't seem necessary at this point? If the enforced wait fixes the test and we don't run back-to-back save/restore, it seems any real human would add the real sleep.

chris-believer (approved):

PR #597, opened 2026-05-11
fix(friendshipper): feedback from zip feature

itscoho:
all four temp-file uses now do flush() → into_temp_path() so the file handle is closed before git reads it, and the to_str() calls are now expect("temp file path is non-UTF-8") instead of bare unwrap().


itscoho:
response to feedback in https://github.com/believer-oss/ethos/pull/593

chris-believer (approved):
😃

PR #602, opened 2026-06-02
Coho/friendshipper fixes

itscoho:


itscoho:
Ok 3 fixes here:
1. fixed our reflog expiry to properly protect our stashes rather than trying to manually manage what refs get messed with
2. fixed a possible initialization error 
3. fixed a bug where if you do git ops when friendshipper is closed, it can show "phantom" modified files on load.  This happens if you have changes, closed friendshipper, did a git reset, and then relaunched friendshipper. these files were coming from the changeset json.

chris-believer (approved):

PR #612, opened 2026-06-24
fix(friendshipper): stop background fetch from triggering interactive git auth

pmarquissee-blv:
## Problem

`GitMaintenanceRunner` runs `git fetch --prune` on a short interval (5s) while the app is open. Each fetch shells out to `git`, which invokes the configured credential helper. When the stored credential is stale, the **background** fetch was free to launch an interactive credential prompt, and concurrent git processes could each spawn their own helper (the `git-credential-manager` "death spiral" the code already guards against). The result is repeated auth pop-ups unrelated to any user action.

## Fix

- Add a `skip_interactive_auth` option to `Opts`. When set, the git command runs with `GIT_TERMINAL_PROMPT=false` and `GCM_INTERACTIVE=never`, so it can never launch an interactive prompt.
- Apply it to the background fetch loop and to `run_maintenance`. A stale credential now fails quietly (and is logged via the existing `warn!`); re-authentication happens on the next user-initiated operation instead of from a background timer.
- Raise the background fetch interval from 5s to 30s to reduce process churn.

No change to which credential helper is used or to auth requirements. User-initiated pull/push still prompt as before.

## Testing

- `cargo check` clean for `ethos-core` and `friendshipper`.
- Pre-commit hook passes: `cargo +nightly clippy --workspace --all-targets --all-features -- -D warnings` and `cargo fmt --all --check`.

Suggested manual check: with a stale `github.com` credential, confirm an idle app no longer launches a login dialog (background fetch fails quietly and logs a warn), while a user-initiated pull/push still prompts and restores the credential.

### DIFF
--- a/core/src/clients/git.rs
+++ b/core/src/clients/git.rs
@@ -106,6 +106,10 @@ pub struct Opts<'a> {
     pub return_complete_error: bool,
     pub lfs_mode: LfsMode,
     pub skip_notify_frontend: bool,
+    // When set, the git command runs with interactive credential prompts
+    // disabled. Use for background/non-user-initiated operations so a stale
+    // credential fails quietly instead of spawning a login window.
+    pub skip_interactive_auth: bool,
 }
 
 #[derive(Clone, Debug, Default)]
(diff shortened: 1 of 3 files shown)


pmarquissee-blv:
Status: a debug build off this branch is out for manual validation with an affected user. What we're confirming: with the app idle, the background fetch no longer triggers GitHub credential prompts; if the stored credential is already stale, there's a single expected prompt on the next user-initiated pull/push, after which it stays quiet. Will update once that's confirmed.

PR #613, opened 2026-06-25
fix(git): serialize subprocesses, seed PAT into git credentials, self-heal auth

itscoho:
## Summary

Friendshipper spawns many `git` subprocesses against a single repo and credential store — foreground ops, a background fetch/maintenance loop, the file-watcher status refresh, and direct handlers. Several ran concurrently, which stomped the Git Credential Manager (GCM) credential (spurious "Connect to GitHub" / re-auth popups unrelated to any user action) and risked `.git` index/ref/packfile collisions.

This branch serializes all app git, keeps background/ambient git non-interactive, and routes the in-app PAT into git's own credential helper so auth is stable and self-healing.

## What this delivers

- **All git subprocesses are serialized.** A process-wide lock is held for the lifetime of every `git` child, acquired at the single command chokepoint plus the few raw-spawn sites (`head_commit`, temp-index ops, `configure_global`, submit's `check-attr`). The app never has two git processes in flight at once. The lock is non-reentrant and ordered consistently with the existing fetch-coalescing lock.

- **A stalled git process can't wedge the app.** An idle (no-output) timeout kills a process that has gone silent (stuck prompt, dead connection) so the lock is released — but never kills a healthy long-running op (clone / fetch / LFS pull) that's still streaming progress.

- **Background and ambient git never prompt.** The fetch + maintenance loop and the status-refresh network calls (`git lfs locks --verify`, upstream `ls-remote`) run with interactive auth disabled, so a stale credential fails quietly instead of popping a login window; re-auth happens only on a deliberate user pull/push. A lock-verify failure preserves the last-known lock state instead of failing the whole status refresh.

- **The PAT actually feeds git.** The stored PAT (previously used only for the GitHub HTTP API) is written into git's credential helper via `git credential approve` over stdin, never on the command line or in logs. Any stale host-level credential is evicted first so the fresh PAT resolves for a bare-host fetch. Seeded at startup and on PAT save.

- **Auth is self-healing and observable.** A background fetch that fails with an auth error auto-repopulates the credential from the stored PAT (recovers within a fetch tick, no restart), with a cooldown so a genuinely-rejected token doesn't churn. Lock contention is logged with the operation that's holding the lock.

## Scope

`core/src/clients/git.rs`, `core/src/clients/git_maintenance_runner.rs`; `friendshipper` `status.rs`, `server.rs`, `config/router.rs`, `gh/submit.rs`. Includes #612.

## Testing

- `cargo clippy --all-targets` / `cargo fmt --check` clean; workspace builds across all three crates.
- Unit tests: credential seed/evict round-trip via `git credential fill` (isolated from the real helper), `is_auth_error` auth-vs-network classification, and the existing snapshot/restore suite (which exercises the multi-acquire lock path).
- Validated against live Friendshipper logs: background fetches and LFS-lock checks succeed with no popups, and a captured 401 recovered via automatic re-seed.

## Reviewer notes (judgment calls)

- Startup seeding **unconditionally** evicts and replaces git's `github.com` credential from the app PAT — intended under a PAT-first policy, but a stale app PAT could regress a working credential.
- The idle-timeout backstop covers the command chokepoint; the local raw-spawn sites hold the lock without it (all are local/fast).
- `GIT_PROCESS_LOCK` is process-local — two overlapping Friendshipper instances could still race the credential store; mitigated by the startup process check and auto-repopulate, not fully prevented.
- Cannot fix a genuinely invalid PAT (e.g. lapsed SSO authorization); re-seeding the same rejected token can't help, but the logs make that case identifiable.

### DIFF
--- a/core/src/clients/git.rs
+++ b/core/src/clients/git.rs
@@ -215,16 +215,28 @@ pub fn parse_bool_string(bool_str: &str) -> anyhow::Result<bool> {
     bail!("Unable to parse string")
 }
 
-/// Wall-clock backstop for a single `git` subprocess. Because every git
-/// process now holds [`GIT_PROCESS_LOCK`] for its whole lifetime, a hung
-/// process — a stuck credential prompt, or a dead connection that never makes
-/// progress — would otherwise hold the lock forever and wedge *every* git
-/// operation in the app. The runner kills such a process so the lock is
-/// released and callers get an error. Set generously so it never trips a
-/// legitimately long op (large fetch, gc/repack). NOTE: a fresh full clone of
-/// a very large repo over a slow link can legitimately exceed this — bump it
-/// or special-case `clone` if that ever bites.
-const GIT_PROCESS_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(30 * 60);
+/// Idle (no-progress) backstop for a single `git` subprocess. Because every git
+/// process now holds [`GIT_PROCESS_LOCK`] for its whole lifetime, a hung process
+/// — a stuck credential prompt, or a dead connection that never makes progress —
+/// would otherwise hold the lock forever and wedge *every* git operation in the
+/// app. The runner kills such a process so the lock is released and callers get
+/// an error.
+///
+/// This is deliberately an *idle* timeout, not a wall-clock one: a hung process
+/// is distinguished from a slow-but-healthy one (a large clone / fetch / LFS
+/// pull streaming progress over a slow link) by the *absence of output*, not by
+/// total elapsed time. We reset the clock on every line the child writes, so a
+/// command that keeps making progress is never killed no matter how long it
+/// runs, while one that goes completely silent for this long is treated as hung.
+/// Generous on purpose — only true silence should trip it; it can be lowered now
+/// that progressing ops are safe.
+const GIT_PROCESS_IDLE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(30 * 60);
+
+/// How often the idle watchdog wakes to compare now against the child's last
+/// output. Bounds how long past [`GIT_PROCESS_IDLE_TIMEOUT`] a hung process can
+/// linger before it's killed; small relative to the timeout, large enough to be
+/// negligible overhead.
+const GIT_PROCESS_IDLE_CHECK: std::time::Duration = std::time::Duration::from_secs(60);
 
 /// Acquire the process-wide git serialization guard. Hold the returned guard
 /// for the full lifetime of a `git` subprocess (spawn through wait). See
@@ -2039,11 +2051,20 @@ impl Git {
         let out_lines: Arc<RwLock<Vec<String>>> = Arc::new(RwLock::new(vec![]));
         let err_lines: Arc<RwLock<Vec<String>>> = Arc::new(RwLock::new(vec![]));
 
+        // Tracks the last time the child produced any output, so the idle
+        // watchdog below can tell a hung process from a slow-but-progressing
+        // one. Both reader tasks bump it on every line.
+        let last_activity: Arc<RwLock<std::time::Instant>> =
+            Arc::new(RwLock::new(std::time::Instant::now()));
+
         let out_lines_thread = out_lines.clone();
+        let out_activity = last_activity.clone();
         let is_collecting_out_lines = output.is_some();
         let should_log_stdout = opts.should_log_stdout;
         let out_handle = tokio::spawn(async move {
             while let Some(line) = out_reader.next_line().await.unwrap() {
+                *out_activity.write() = std::time::Instant::now();
+
                 // Shorten any line starting with "Updating files". Git currently sends us a huge
                 // wall of text with all the individual percentage updates, instead of one line
                 // per update.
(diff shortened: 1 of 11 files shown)


itscoho:
@pmarquissee-blv heads up I integrated your PR into this

core/src/clients/git.rs
```diff
@@ -33,6 +33,34 @@ lazy_static! {
...
+    /// index / refs / packfiles can collide. Every subprocess spawn site holds
+    /// this guard for the lifetime of its child, so the app never has two `git`
+    /// processes in flight at once. External git processes (a user's terminal,
+    /// the engine) are out of scope — `wait_for_lock` mitigates index.lock
+    /// contention with those.
+    ///
+    /// INVARIANT — this lock is NOT re-entrant (tokio's Mutex deadlocks on a
+    /// second acquire from the same task). Every acquisition site is a leaf
+    /// that spawns exactly one subprocess and never calls another acquiring
+    /// function while holding the guard. `GIT_FETCH_LOCK` is always taken
```
chris-believer, line 50: Is GIT_FETCH_LOCK unnecessary now?

core/src/clients/git.rs
```diff
@@ -1638,6 +1800,102 @@ impl Git {
...
+            "Refreshing git credential for {host} from the saved PAT (username: {username}): \
+             evicting any stale host-level entry, then re-seeding the credential helper so \
+             background fetch/pull/push/lfs authenticate without an interactive login."
+        );
+
+        // Spawns git processes (and the helper), so hold the process lock for
+        // the whole reject + approve pair like every other spawn site.
+        let _git_lock = acquire_git_process_lock("git credential (reject + approve)").await;
+
+        // Step 1 — evict. Best-effort: `reject` is a no-op when nothing is
```
chris-believer: A little worried about this break-then-make flow. It doesn't look like it's called often, but it feels like a better path here would be to use fill to check and only `approve` a new one if it doesn't match. Since we don't fully _own_ this data, a hiccup on the approve or something else reading the entry with bad timing would best be avoided.

friendshipper/src-tauri/src/repo/operations/status.rs
```diff
@@ -178,8 +177,25 @@ where
 
         {
             status.lock_user.clone_from(&self.github_username);
-            status.locks_ours = locks.ours;
-            status.locks_theirs = locks.theirs;
+            match locks {
```
chris-believer, line 180: Does taking a lock on an asset still trigger an interactive prompt if we have stale lock information? IOW, do we rely on `status.locks_{ours,theirs}` being authoritative while taking a lock?

core/src/clients/git.rs
itscoho: Not unnecessary, but it does something narrower than GIT_PROCESS_LOCK: it coalesces fetches. fetch() uses try_lock — if a fetch is already in flight, the second caller skips its own fetch and returns once the first finishes, instead of running a redundant second one. GIT_PROCESS_LOCK would only serialize those two (run them back-to-back), not dedupe them. So a foreground refresh arriving during the 30s background fetch collapses to one network round-trip.

That said, it's an optimization, not correctness  so we can remove if you feel strongly about it

core/src/clients/git.rs
itscoho: good point, I'll fix this one

friendshipper/src-tauri/src/repo/operations/status.rs
itscoho: So LockOp uses the PAT directly with the API so no prompt there.

Claude had this to say about the cached locks and makes sense

>  Do we rely on cached locks being authoritative? Only as an advisory client-side filter/pre-check — locks_ours filters redundant paths, locks_theirs gates the handler. The LFS server is authoritative on the actual lock: stale cache can let a request through that the server then cleanly rejects, but it can't grant an incorrect lock. And note this is strictly better than before — a verify_locks failure used to fail the entire status refresh, so you'd have had no lock info at all.


The lock UI can show stale data if "verify_locks" is failing and we could show some sort of indicator in that case 🤔

core/src/clients/git.rs
itscoho: Ok should be better now :)

itscoho:
@chris-believer added a pinning to a specific user name to ensure that we play nice with multiple accounts on a machine

itscoho:
in addition, added this checkbox to disable credential seeding if people don't want it (though it defaults to on)
image

core/src/clients/git.rs
```diff
@@ -1800,35 +1800,123 @@ impl Git {
...
+    /// and approve steps of [`Self::store_credential`].
+    fn build_credential_command(&self, action: &str) -> anyhow::Result<Command> {
+        self.build_raw_command(&["credential", action])
+    }
+
+    /// Read a single value from the repo's **local** git config; `None` when
+    /// the key is unset (or unreadable). Caller must hold the git process lock.
+    async fn local_config_get(&self, key: &str) -> Option<String> {
+        let out = self
+            .build_raw_command(&["config", "--local", "--get", key])
```
chris-believer, line 1859: Pedantic, but if we're creating a `build_credential_command` should we also do a `build_config_command`, and if not rewrite the three uses of `build_credential_command` into `build_raw_command`?

core/src/clients/git.rs
```diff
@@ -1877,14 +2433,39 @@ impl Git {
...
+            match tokio::time::timeout(GIT_PROCESS_IDLE_CHECK, git_proc.wait()).await {
+                Ok(wait_result) => break wait_result?,
+                Err(_elapsed) => {
+                    let idle = last_activity.read().elapsed();
+                    if idle >= GIT_PROCESS_IDLE_TIMEOUT {
+                        warn!(
+                            "git command made no progress for {:?}; killing stalled process: {}",
+                            idle, git_cmd_str
+                        );
+                        let _ = git_proc.kill().await;
```
pmarquissee-blv, line 2505: this kill path can wedge every git op in the app on Windows.

kill() only terminates git.exe. The children it spawned (git-remote-https, the GCM prompt process) survive, and they inherited the stderr pipe's write end. So the reader tasks never see EOF, the out_handle.await / err_handle.await joins below never return, this function never exits, and the GitLockGuard never drops. Every git subprocess in the app queues behind a lock that will never release, until restart.

A stuck GCM credential prompt is the exact scenario this watchdog exists for, and it guarantees a live orphan holding the pipe. So the watchdog defeats itself precisely when it's needed.

Two ways out:
1. Cheap: wrap the two joins in a short tokio::time::timeout and abort() the handles on expiry. ~5 lines, fixes the hang, orphans linger but the app stays alive.
2. Right: put the child in a Windows Job Object with JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE so the whole tree dies together. More work, also fixes the orphan-still-writing-credentials hole.

chris-believer (dismissed):
Love that we're managing creds, and allowing users to opt out if needed. Tiny nit, no worries either way!

core/src/clients/git.rs
itscoho: Went with option 2 — dropped the `build_credential_command` wrapper and the three call sites use `build_raw_command(&["credential", ...])` directly (the config side already has its own `local_config_get/set/unset` helpers). Fixed in aaf7798.

core/src/clients/git.rs
itscoho: Great catch — did both in aaf7798:

1. New `ProcessTreeJob` (Job Object with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`) assigned right after spawn; the watchdog now terminates the job so the whole tree (git-remote-https, GCM prompt) dies and the pipes actually close. Normal exits drop the job before the joins to reap stragglers. Best-effort: if create/assign fails we degrade to plain kill.
2. Cross-platform backstop: both join sites go through a 5s bounded drain that aborts the reader tasks on timeout, so the process lock can never wedge even if the job path fails.

Covered by a test where `cmd` spawns `ping` sharing our stdout pipe — after `terminate()` the pipe must hit EOF, which only happens if the grandchild died too.

itscoho: good response robo-me

PR #619, opened 2026-08-07
chore(misc): add claude code park and workflow skills

rudoi:
## Summary

Adds optional [Claude Code](https://claude.com/claude-code) tooling under `.claude/`. Nothing here is required to build, test, or contribute — the normal `cargo` and `yarn` workflows are untouched.

- **`/park`** — resets a worktree to latest `origin/main`, parking the current branch on a local-only branch. It refuses to discard uncommitted changes or unpushed commits, always stashes before a forced reset (and prints the stash ref), detects branches that were already squash-merged via `git cherry`, and retries on git lock contention so several worktrees can park concurrently.
- **`/workflow`** — advances a 10-step workflow: plan → refine → fracture → singletonize → loss check → sequence → implement → test → stage → review gate. It delegates proposal review and work breakdown to two instruction documents.

The instruction docs encode this repo's actual conventions so an agent doesn't have to guess: the workspace layout, the CI-equivalent commands, the `pre-commit`/`commit-msg` hooks (including that the hook's nightly workspace clippy is stricter than CI), the commitlint type/scope enums, and a branch convention that prefixes agent-driven branches with `llm/` to keep them distinguishable from human ones.

## `.gitignore` change

`.claude` was ignored wholesale. It now ignores `.claude/*` and re-includes only the four shared paths:

```
.claude/*
!.claude/README.md
!.claude/instructions/
!.claude/scripts/
!.claude/skills/
```

So the shared tooling is tracked while per-user state — `.claude/workflow/` (in-progress plan documents and workflow state) and `settings.local.json` — stays ignored. Verified both directions with `git check-ignore`.

## Unrelated lint fix, included because it blocked committing

Two `chunks_exact` calls are rewritten to `as_chunks`:

- `friendshipper/src-tauri/src/repo/operations/gh/submit.rs`
- `friendshipper/src-tauri/src/system/unreal.rs`

Recent nightly clippy added `chunks_exact_to_as_chunks`, and the `pre-commit` hook runs `cargo +nightly clippy --workspace --all-targets --all-features -- -D warnings`, so this lint blocked **every** local commit regardless of what was being staged. Both rewrites drop the trailing remainder exactly as `chunks_exact` did, so behavior is unchanged; in `submit.rs` the fixed-size array pattern is now infallible, so its `unreachable!` arm is gone.

Worth noting separately: because the hook pins `+nightly` with `-D warnings` and there is no toolchain file, any new nightly lint becomes a commit blocker for everyone, while a nightly older than the dependency floor (currently rustc 1.91.1) fails to build at all. That's a narrow window, and it may be worth pinning a toolchain.

## Test plan

Verified locally on Windows, capturing each command's own exit code:

- [x] `cargo +nightly clippy --workspace --all-targets --all-features -- -D warnings` (the hook's check) — clean
- [x] `cargo clippy --all-features -p ethos-core -- -D warnings` — clean
- [x] `cargo clippy --all-features -p friendshipper -- -D warnings` — clean
- [x] `cargo fmt --all --check` — clean
- [x] `cargo test -p ethos-core` — 34 passed, 5 ignored
- [x] `cargo test --bin friendshipper` — passed
- [x] `/park`'s safety path exercised: it correctly refused to run against a dirty tree

Not verified locally, flagged honestly:

- The `submit.rs` path has no test covering it, so that rewrite rests on the compiler and on `as_chunks().0` being equivalent to `chunks_exact` — not on an exercised test.
- `unreal.rs` is Windows-only, so Linux CI won't compile that hunk.
- `as_chunks` requires rustc >= 1.88; CI resolves `stable`, which is well past that.

### DIFF
--- a/.claude/README.md
+++ b/.claude/README.md
@@ -1,0 +1,88 @@
+# Claude Code Tooling
+
+Optional tooling for contributors who use [Claude Code](https://claude.com/claude-code). Nothing here is required to build, test, or contribute to ethos — the normal `cargo` and `yarn` workflows are unaffected.
+
+The tooling below is shared; everything else under `.claude/` is per-user and ignored:
+
+```
+.claude/                         # tracked:
+  README.md                      #   this file
+  skills/park/SKILL.md           #   /park — reset this worktree to latest origin/main
+  skills/workflow/SKILL.md       #   /workflow — advance the workflow one step
+  instructions/
+    DoNextWorkflowStep.md        #   the 10-step workflow orchestrator
+    RefineProposal.md            #   structured proposal review (Stage 2A)
+    BreakDownWorkProposal.md     #   proposal -> ordered implementation stages (Stage 5)
+  scripts/
+    Reset-WorktreeToMain.ps1     #   the mechanics behind /park (PowerShell)
+
+  workflow/                      # ignored: per-user workflow state, created on demand
+    <work-slug>/                 #   plan docs, NNN stage files, WORKFLOW_STATE.json
+    archive/<work-slug>/         #   design docs preserved after staging
+  settings.local.json            # ignored: per-user settings
+```
+
+`.gitignore` ignores `.claude/*` and then re-includes only the tracked paths above, so personal settings and in-progress workflow state never land in the repository.
+
+## Skills
+
+| Skill | What it does |
+|---|---|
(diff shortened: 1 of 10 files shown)


itscoho (approved):
Good stuff 👍

PR #624, opened 2026-09-09
Llm/itscoho/promotion destinations

itscoho:
* Added new promotion flow options to go to steam and support promotions without backend deploys if desired
* Added new "Active Builds" Modal and Button to know what is currently deployed
* Added new "Promoted" icon ⬆️ with hover text describing where build is promoted to builds list

### DIFF
--- a/core/src/types/argo/workflow.rs
+++ b/core/src/types/argo/workflow.rs
@@ -44,9 +44,69 @@ pub struct WorkflowTemplateRef {
 #[derive(Deserialize, Serialize, Clone, Debug, Default, JsonSchema)]
 pub struct CreatePromoteBuildWorkflowRequest {
     pub commit: String,
-    pub shard: Option<String>,
+    /// Backend environment for the deploy. The serde name stays `shard`
+    /// because it is the Argo template's parameter name and the wire field
+    /// shared with friendshipper-server.
+    #[serde(rename = "shard")]
+    pub backend_environment: Option<String>,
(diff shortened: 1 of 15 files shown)


core/src/types/argo/workflow.rs
```diff
@@ -44,9 +44,71 @@ pub struct WorkflowTemplateRef {
...
+}
+
+impl CreatePromoteBuildWorkflowRequest {
+    /// Assemble the Argo workflow parameters for this promotion request.
+    ///
+    /// `backend_environment` is deliberately three-state and the distinction is
+    /// load-bearing: `None` omits the parameter entirely so the workflow
+    /// template's own default applies, `Some("")` emits it empty to suppress the
+    /// backend deploy, and `Some(v)` emits the configured environment. Never
+    /// normalize `None` into `Some("")`.
```
chris-believer: I had to read this a couple of times to understand it, so might be worth calling this out. 😄 
https://github.com/BelieverCo/infra/blob/flux/flux/argo-pipelines/builddev-v2/templates/promote-build.yaml#L75-L77
None and Some("") are equivalent, the default for the workflow is "" and I have no reason to believe it would ever make sense to change. That's a common pattern in the argo workflows, optional parameters are not possible, so you typically default to an empty string and then test for it.

itscoho:
Chris made a couple great callouts, fixing now

chris-believer (approved):

PR #631, opened 2026-09-17
fix(friendshipper): clear read-only before overwriting on zip import

itscoho:
Extracting over an unlocked lockable asset failed with "Access is denied" on Windows and aborted the import. Clear the attribute first and leave the file writable, matching what snapshot restore leaves behind.


core/src/fs.rs
```diff
@@ -44,3 +44,84 @@ pub fn check_dir(path: PathBuf) -> Option<FsEntry> {
...
+///
+/// Working-tree files land read-only through several routes — git-lfs marks
+/// lockable files read-only when the user doesn't hold the lock (see the
+/// `GIT_LFS_SET_LOCKABLE_READONLY` handling in [`crate::operations`]), and
+/// content copied in from other tooling often carries the attribute too. On
+/// Windows, opening such a file for writing fails with "Access is denied".
+///
+/// A missing file is not an error, so this can be called as a precondition
+/// before writing a path that may not exist yet.
+pub fn clear_readonly(path: &Path) -> std::io::Result<()> {
```
chris-believer, line 59: Not a blocker at all, just musing, but feels like we should have a mirror set_readonly fn here that replaces the giant stanza in operations.rs.

chris-believer (approved):
Nice cleanup!

